PR #8, opened 2018-01-22
Deployment helpers + Service JSON script 

ajvb:
Building off of @g-k's scripts, this PR gets pytest-services ready to be deployed. Also adds `make awsci` command to allow for an AWS-specific deployment to start.


Makefile
```diff
@@ -5,19 +5,26 @@ ENTER_VENV := source venv/bin/activate
...
+
+clean-aws-ci:
+	venv/bin/pytest --cache-clear aws/ --aws-profiles $(AWS_PROFILE) --json=results-$(AWS_PROFILE)-$(TODAY).json
 
 venv:
 	python3 -m venv venv
 
 install: venv
-	$(ENTER_VENV) && pip3 install -r requirements.txt
+	venv/bin/pip3 install -r requirements.txt
```
g-k: 👍 this is cleaner

Makefile
```diff
@@ -5,19 +5,26 @@ ENTER_VENV := source venv/bin/activate
 
 .DEFAULT_GOAL := all
 
+AWS_PROFILE := default
```
g-k: If we run these the same way we did before, on the ops jenkins with a read only instance level AWS profile, then it'll likely throw an exception for a profile named `default` not existing (since there isn't one in `.aws/credentials`). We ended up switching to `None` (the type not the string `'None'`) to fix the issue.

Something to keep in mind.

conftest.py
```diff
@@ -104,11 +105,13 @@ def pytest_runtest_makereport(item, call):
...
         fixtures = {fixture_name: item.funcargs[fixture_name] for fixture_name in item.fixturenames if fixture_name != 'request'}
 
+
         # add json metadata
         report.test_metadata = dict(
             outcome=outcome,  # 'passed', 'failed', 'skipped', 'xfailed', 'xskipped', or 'errored'
             reason=reason,
-            unparametrized_name=item.originalname or item.name,
+            unparametrized_name=item.originalname,
+            parametrized_name=item.name,
```
g-k: 👍 explicit is better than implicit ...

g-k:
Updated Travis CI to only run tests when its config file is present ...

Makefile
ajvb: Absolutely! I changed pytest_wrapper.sh to run `make awsci AWS_PROFILE=$AWS_PROFILE`, so each job should be passing it in.

PR #9, opened 2018-01-23
Encrypted EBS Volumes and Admin Users MFA tests

ajvb:
This PR adds two tests:

- test_ec2_ebs_volume_encrypted
- test_iam_admin_user_without_mfa

We will want to eventually change `test_iam_admin_user_without_mfa` to test for all users that can log in, but for now this is a nice scoping.

### Discussion

If you look at the `user_is_admin` function, there is a comment block above it:
```
# FIXME
# Substring matching is _not_ enough of a check, but works for testing.
# The truth is that we probably shouldn't depend too much on the concept
# of an "admin" in AWS, since that's not how the ACL's really work. We
# should probably move towards concepts like "has write access", "can
# read secrets", etc.
```

Something I'd like to discuss is whether attempting to scope to the concept of an "Admin" via policies is ok for now or if this should get more flushed out at this point.

### DIFF
--- a/aws/client.py
+++ b/aws/client.py
@@ -209,8 +209,8 @@ class BotocoreClient:
                 keyed_result = result[key]
                 if isinstance(keyed_result, list):
                     for item in keyed_result:
-                        # TODO: Is there a bug in usage (IAM resources) causing it
-                        # to sometimes be a string?
+                        # Added for IAM inline policies call, as it
+                        # returns a list of strings.
                         if not isinstance(item, str):
                             item['__pytest_meta'] = result['__pytest_meta']
                 elif isinstance(keyed_result, dict):
--- a/aws/client.py
+++ b/aws/client.py
@@ -202,7 +202,10 @@ class BotocoreClient:
                 keyed_result = result[key]
                 if isinstance(keyed_result, list):
                     for item in keyed_result:
-                        item['__pytest_meta'] = result['__pytest_meta']
+                        # TODO: Is there a bug in usage (IAM resources) causing it
+                        # to sometimes be a string?
+                        if not isinstance(item, str):
+                            item['__pytest_meta'] = result['__pytest_meta']
                 elif isinstance(keyed_result, dict):
                     keyed_result['__pytest_meta'] = result['__pytest_meta']
 
--- a/aws/client.py
+++ b/aws/client.py
@@ -11,6 +11,9 @@ import botocore.exceptions
 import botocore.session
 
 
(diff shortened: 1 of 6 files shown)


ajvb:
@g-k Oh, also, the concept of `regions` within the `BotocoreClient` as implemented [here](https://github.com/mozilla-services/pytest-services/blob/master/aws/client.py#L105) isn't what we will want for `IAM`. Where do you think would be the best place to insert logic around services that don't use regions like other services?

g-k:
> attempting to scope to the concept of an "Admin" via policies is ok for now or if this should get more flushed out at this point

I think it's fine for now, but I'd open an issue for defining what Admin means (+1 being able to list permissions eventually).

We might want to pass these things a config files, since other orgs might have different definitions of user and admin.

g-k:
> Where do you think would be the best place to insert logic around services that don't use regions like other services?

Huh, TIL.

Maybe:

* throw a conditional in [get_client](https://github.com/mozilla-services/pytest-services/blob/master/aws/client.py#L28)
* add a `get_iam_resource` method or `get_resource_without_regions` that pulls the resource for all profiles

I'm open to other approaches too and opened a couple issue around fixing the client too (#11, #12).

aws/client.py
```diff
@@ -202,7 +202,10 @@ def extract_key(self, key, default=None):
                 keyed_result = result[key]
                 if isinstance(keyed_result, list):
                     for item in keyed_result:
-                        item['__pytest_meta'] = result['__pytest_meta']
+                        # TODO: Is there a bug in usage (IAM resources) causing it
+                        # to sometimes be a string?
```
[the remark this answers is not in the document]
ajvb: Oh I think I know what causes this. The "inline" IAM policy methods return a list of policy names rather than a list of dictionaries. (http://botocore.readthedocs.io/en/latest/reference/services/iam.html#IAM.Client.list_user_policies)

ajvb:
> throw a conditional in get_client

This seemed like the best solution to me, especially since there are very few services that don't utilize regions like everyone else.

aws/client.py
```diff
@@ -163,6 +166,10 @@ def get(self,
             profiles=None,
             regions=None,
             result_from_error=None):
+
+        if service_name in SERVICES_WITHOUT_REGIONS:
+            regions = ["us-east-1"]
+
```
ajvb: @g-k Thoughts on this?

g-k: Not great, but we're hardcoding stuff anyway so why not.

ajvb: Haha totally.

Issue #13, opened 2018-01-24
Run doctests in CI


g-k:
This is a nice intermediate step that doesn't require the full refactor in #3 or running against cached data.

g-k:
Possible solution for running doctest in the data collection files is to run in CI against anonymous cached data or a test AWS account with `--doctest-modules` or `--doctest-glob` https://docs.pytest.org/en/latest/doctest.html

PR #15, opened 2018-01-24
Move RDS helpers to helpers.py + add 'make doctests'

ajvb:
Prep work for #13 

Moving them all to `helpers.py` was to not have to deal with doctest executing pytest decorators. Would get errors around `botocore_client` being a `None type` due to not having pytest set it up. Don't really want those decorators executing anyways.


g-k:
We should update the readme to document the new and improved helpers.py convention too.

PR #21, opened 2018-01-29
Add security group checks

g-k:
refs: https://github.com/mozilla-services/foxsec/issues/381

Adds checks for EC2 and Redshift security groups and renames the old ec2 check and moves its test predicate to a helpers file.

r? @ajvb


aws/redshift/resources.py
```diff
@@ -0,0 +1,12 @@
+
+
+from conftest import botocore_client
+
+
+def redshift_cluster_security_groups():
+    "http://botocore.readthedocs.io/en/latest/reference/services/rds.html#RDS.Client.describe_db_instances"
```
ajvb: wrong doc string?

g-k: yeah, good catch!

PR #23, opened 2018-01-30
aws: Add ec2 instance required tag name test

g-k:
refs: https://github.com/mozilla-services/foxsec/issues/381

Adds a test for required EC2 instance tags.

example usage and output:

```console
pytest --ignore pagerduty/ --ignore aws/s3 --ignore aws/rds --ignore aws/iam -s --aws-profiles cloudservices-aws-stage --aws-regions us-east-1 --aws-debug-calls --aws-require-tag Name --aws-require-tag Type --aws-require-tag App --aws-require-tag Stack -k test_ec2_instance_has_required_tags
...
_____________________________________________ test_ec2_instance_has_required_tags[i-0e537fff] ______________________________________________

ec2_instance = {'AmiLaunchIndex': 0, 'Architecture': 'x86_64', 'BlockDeviceMappings': [{'DeviceName': '/dev/sda1', 'Ebs': {'AttachTim...11:53+00:00', 'DeleteOnTermination': True, 'Status': 'attached', 'VolumeId': 'vol-11d6ffff'}}], 'ClientToken': '', ...}
required_tag_names = frozenset({'App', 'Name', 'Stack', 'Type'})

    @pytest.mark.ec2
    @pytest.mark.parametrize(
        'ec2_instance',
        ec2_instances(),
        ids=ec2_instance_test_id)
    def test_ec2_instance_has_required_tags(ec2_instance, required_tag_names):
        """
        Checks that all EC2 instances have the tags with the required names.
    
        Does not not check tag values.
        """
        instance_tag_names = set(tag['Key'] for tag in ec2_instance['Tags'])
    
        # set difference to find required tags not on instance
        missing_tag_names = required_tag_names - instance_tag_names
>       assert not missing_tag_names, \
          "EC2 Instance {0[InstanceId]} missing required tags {1!r}".format(ec2_instance, missing_tag_names)
E       AssertionError: EC2 Instance i-0e537fff missing required tags frozenset({'Stack'})
E       assert not frozenset({'Stack'})

aws/ec2/test_ec2_instance_has_required_tags.py:27: AssertionError
```

r? @ajvb

### DIFF
--- a/aws/ec2/test_ec2_instance_has_required_tags.py
+++ b/aws/ec2/test_ec2_instance_has_required_tags.py
@@ -19,8 +19,18 @@ def test_ec2_instance_has_required_tags(ec2_instance, required_tag_names):
     Checks that all EC2 instances have the tags with the required names.
 
     Does not not check tag values.
+
+    >>> test_ec2_instance_has_required_tags({'Tags': [{'Key': 'Name'}]}, frozenset(['Name']))
+
+    >>> test_ec2_instance_has_required_tags({
+    ... 'InstanceId': 'iid', 'Tags': [{'Key': 'Bar'}]}, frozenset(['Name']))
+    Traceback (most recent call last):
+    ...
+    AssertionError: EC2 Instance iid missing required tags frozenset({'Name'})
+    assert not frozenset({'Name'})
     """
-    instance_tag_names = set(tag['Key'] for tag in ec2_instance['Tags'])
+    tags = ec2_instance.get('Tags', [])
+    instance_tag_names = set(tag['Key'] for tag in tags if 'Key' in tag)
 
     # set difference to find required tags not on instance
     missing_tag_names = required_tag_names - instance_tag_names
--- a/aws/ec2/test_ec2_instance_has_required_tags.py
+++ b/aws/ec2/test_ec2_instance_has_required_tags.py
@@ -6,7 +6,7 @@ from aws.ec2.resources import ec2_instances
 
 @pytest.fixture
 def required_tag_names(pytestconfig):
-    return frozenset(pytestconfig.getoption('--aws-require-tag'))
+    return frozenset(pytestconfig.getoption('--aws-require-tags'))
 
 
 @pytest.mark.ec2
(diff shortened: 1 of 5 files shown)


aws/ec2/test_ec2_instance_has_required_tags.py
```diff
@@ -0,0 +1,28 @@
+import pytest
+
+from aws.ec2.helpers import ec2_instance_test_id
+from aws.ec2.resources import ec2_instances
+
+
+@pytest.fixture
+def required_tag_names(pytestconfig):
+    return frozenset(pytestconfig.getoption('--aws-require-tag'))
```
g-k: We could move this to conftest.py if other tests need it.

Alternatively the test could use `pytestconfig` directly, but having it as as separate fixture might make caching it easier.

conftest.py
```diff
@@ -30,6 +30,11 @@ def pytest_addoption(parser):
                      action='store_true',
                      help='Log whether AWS API calls hit the cache. Requires -s')
 
+    parser.addoption('--aws-require-tag',
```
ajvb: NB: Would it be worth following the pattern of having the options be plural? 

Changing it to be:
```
--aws-require-tags Name,Type,App,Stack
```

g-k: Yeah, they're inconsistent. I was thinking of making the other aws args (profiles and regions) singular, since the parser chokes on CSVs.

I'll make it plural and open an issue to discuss how we want to handle those args.

ajvb: Sounds good.

I've got a simple fix locally, which looks like this:

```
diff --git a/conftest.py b/conftest.py
index 3b1fe43..0247b21 100644
--- a/conftest.py
+++ b/conftest.py
@@ -30,6 +30,14 @@ def pytest_addoption(parser):
                      action='store_true',
                      help='Log whether AWS API calls hit the cache. Requires -s')
 
+def parse_opt(opt):
+    if not len(opt):
+        return None
+
+    if ',' in opt[0]:
+        return opt[0].split(',')
+
+    return opt
 
 def pytest_configure(config):
     global botocore_client
@@ -37,13 +45,7 @@ def pytest_configure(config):
     # monkeypatch cache.set to serialize datetime.datetime's
     patch_cache_set(config)
 
-    profiles, regions = config.getoption('--aws-profiles'), config.getoption('--aws-regions')
-
-    if not len(profiles):
-        profiles = None
-
-    if not len(regions):
-        regions = None
+    profiles, regions = parse_opt(config.getoption('--aws-profiles')), parse_opt(config.getoption('--aws-regions'))
```

PR #41, opened 2018-01-31
use nargs for multi value args

g-k:
refs https://github.com/mozilla-services/pytest-services/issues/31

I think this is slightly more ergonomic.

r? @ajvb

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ And results in a severity and severity marker being included in the
 json metadata:
 
 ```console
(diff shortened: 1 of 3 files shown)


g-k:
I think in practice we shouldn't ever have commas in region, profile, or tag names. 

@hwine do you have an opinion re: CSV or space delimited args?

g-k:
Kind of ambivalent about this, but it's what I was originally going for so I'm going to land this.

We can bikeshed further if it makes config files weird.

PR #45, opened 2018-01-31
Fix type error for when an opt is not passed

ajvb:
Fixes `TypeError: object of type 'NoneType' has no len()` when either profiles/regions opts are missing.

r? @g-k


g-k:
Merging to unbork local dev.

PR #55, opened 2018-02-01
Support AWS_SESSION_TOKEN within authentication

ajvb:
This allows a user to run pytest-services as a user with MFA.

This implementation is alright, not a huge fan. The addition
of access key and secret key is caused by a conditional within
botocore's create_client method:
  https://github.com/boto/botocore/blob/develop/botocore/session.py#L838-L842

r? @g-k

### DIFF
--- a/aws/client.py
+++ b/aws/client.py
@@ -22,6 +22,10 @@ def get_session(profile=None):
     # we proceed.
     if 'AWS_PROFILE' in os.environ and os.environ['AWS_PROFILE'] != profile:
         del os.environ['AWS_PROFILE']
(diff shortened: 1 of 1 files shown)


ajvb:
Not needed. Was using `--aws-profiles` when I should have not been.

g-k:
@ajvb do you think we should warn if `--aws-profiles` and `AWS_SESSION_TOKEN` are present?

ajvb:
@g-k I was thinking about this last night. I'd say yes or warn if `--aws-profiles` and `$AWS_PROFILE` don't match.

PR #57, opened 2018-02-02
Small fixes for pytest_wrapper.sh/make install

ajvb:
* `make install` now sources venv before pip installing
* Have pytest_wrapper.sh source the venv itself

r? @g-k

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ flake8: check_venv
 	flake8 --max-line-length 120 $(shell find . -name '*.py' -not -path "./venv/*")
 
 install: venv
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -29,7 +29,7 @@ flake8: check_venv
 	flake8 --max-line-length 120 $(shell find . -name '*.py' -not -path "./venv/*")
 
 install: venv
-	pip install -r requirements.txt
+	( . venv/bin/activate; pip install -r requirements.txt )
```
g-k: This should fail if activation fails (so `&&` instead of `;`) right?

ajvb: absolutely.

PR #62, opened 2018-02-02
Exemptions conf

### LINKED ISSUE #18: Add ability to mark a resource as exempt from a certain test
Add ability to mark a resource as exempt from a certain test along side an optional reason and expiration date. Should include the ability to mark a resource as exempt from more than one test and from all tests.

Would also be nice to add the ability to pull in suppressed checks on resources from trusted advisor (http://botocore.readthedocs.io/en/latest/reference/services/support.html#Support.Client.describe_trusted_advisor_check_result).  

edit (g-k): 

Moved the trusted advisor check to https://github.com/mozilla-services/pytest-services/issues/37

Planning to do this similar to the severity config #28 i.e.

- [ ] add a conf file
- [ ] for each exemption add an xfail marker in `pytest_runtest_setup` with reason, expiration time, and the conf filename and line number
- [ ] add exemptions section to readme

Markers will be serialized and included in the test metadata.
It will need to match on resource and test names.

g-k:
Fixes #18 up to a per test name and ID level. 

The first two commits fix other doctests and other stuff.

r? @ajvb

### DIFF
--- a/exemptions.py
+++ b/exemptions.py
@@ -1,0 +1,138 @@
+
+
+from datetime import datetime
+import warnings
+
+import pytest
+
+
+def parse_conf_file(conf_fd):
+    """Marks tests as xfail based on test name and ID.
+
+    Parses an open exemptions conf file and returns a two level dict with format:
+
+    {<test_name>:
+        {<test_id>: (<exemption line number>, <exemption expiration>, <exemption reason>)
+        ...
+        }
+    ...
+    }
+
+    The config file consists of:
+
+    comment lines that have the format:
+
+    '#'<anything>\n
+
+    and exemption lines with the format:
+
+    <test name><whitespace><test id><whitespace><exemption expiration><whitespace><exemption reason>\n
+
+    where:
+
+    <anything> := is any non newline character
+    <exemption line number> := line number from the exemption config file
+    <exemption expiration> := date in YYYY-MM-DD format the exemption expires
+    <exemption reason> := reason for the test exemption (e.g. S3 bucket contains public data)
+    <test name> := unparametrized test name
+    <test id> := id from the parametrize decorator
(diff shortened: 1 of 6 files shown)


exemptions.py
```diff
@@ -0,0 +1,138 @@
...
+            expiration_date = datetime.strptime(expiration, '%Y-%m-%d')
+        except ValueError:
+            warnings.warn('Line {}: Skipping line with invalid expiration day {!r}'.format(line_number, expiration))
+            continue
+        if expiration_date < datetime.now():
+            warnings.warn('Line {}: Skipping line with expiration day in the past {!r}'.format(line_number, expiration))
+            continue
+
+        if test_name not in rules:
+            rules[test_name] = {}
```
ajvb: NB: Want to use a `defaultdict` for `rules`?

g-k: Yeah, good idea. Updated.

exemptions.py
```diff
@@ -0,0 +1,136 @@
...
+    ... 'test_foo foo-id 2051-01-01 in prod never allow foo for bar-id'
+    ... ))
+    {'test_foo': {'foo-id': (0, '2050-01-01', 'in prod never allow foo for foo-id')}}
+    >>> # UserWarning: Line 1: Skipping line with duplicate test name and ID 'test_foo' 'foo-id'
+
+    Does not check that test name and IDs exist (since names might not
+    be collected and IDs can require an HTTP call).
+    """
+    # dict of test name to severity level
+    rules = defaultdict(lambda: {})
```
ajvb: I think you want `defaultdict(dict)`, or am I missing something?

g-k: They behave the same since it's just looking for a callable, but that's cleaner so I switched to `dict`.

ajvb: oh gotcha, cool cool :)

PR #64, opened 2018-02-02
Added default IAM policy for pytest-services

ajvb:
Policy that can be thrown on a user to allow for running all pytest-services' AWS tests.

r? @g-k 

@g-k Also, do you agree/think that this is the right place to store this?


g-k:
> Also, do you agree/think that this is the right place to store this?

Yeah, there's nothing specific to our accounts in it.

PR #65, opened 2018-02-05
 Further tweaks for CI workflow 

ajvb:
1. Add support for optional severity/exemptions config in
   pytest_wrapper.sh
2. Add '--continue-on-collection-errors' to ci make targets
3. Default skip redshift sec groups test
4. Remove passing in --aws-profiles for make targets
5. Fix serializing fixtures error

In regards to #4 - within our CI workflow, it is significantly
cleaner to use AWS's env var flow than create an AWS credentials
file.

In regards to #5:
Fixtures contain the boto response which has a datetime
object in it, causing "TypeError: datetime.datetime()
is not JSON serializable" errors to be thrown. Since we
are not currently using the fixtures metadata, simply
remove it

r? @g-k

### DIFF
--- a/aws/redshift/test_redshift_security_group_does_not_allow_all_ips_access.py
+++ b/aws/redshift/test_redshift_security_group_does_not_allow_all_ips_access.py
@@ -10,7 +10,7 @@ from aws.redshift.helpers import (
 
 
 @pytest.mark.redshift
-@pytest.mark.xskip(reason="Most don't use Redshift security groups")
+@pytest.mark.xskip(reason="Redshift security groups are only supported on EC2-Classic accounts")
 @pytest.mark.parametrize(
     'security_group',
     redshift_cluster_security_groups(),
--- a/aws/redshift/test_redshift_security_group_does_not_allow_all_ips_access.py
+++ b/aws/redshift/test_redshift_security_group_does_not_allow_all_ips_access.py
@@ -10,6 +10,7 @@ from aws.redshift.helpers import (
 
 
 @pytest.mark.redshift
+@pytest.mark.xskip(reason="Most don't use Redshift security groups")
 @pytest.mark.parametrize(
     'security_group',
     redshift_cluster_security_groups(),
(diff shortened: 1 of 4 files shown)


aws/redshift/test_redshift_security_group_does_not_allow_all_ips_access.py
```diff
@@ -10,6 +10,7 @@
 
 
 @pytest.mark.redshift
+@pytest.mark.xskip(reason="Redshift security groups are only supported on EC2-Classic accounts")
```
g-k: the more I know :)

I wonder if we could detect this somehow

conftest.py
```diff
@@ -168,16 +168,9 @@ def pytest_runtest_makereport(item, call):
             clean_docstring(markers.get('rationale')['args'][0])
         description = item._obj.__doc__ and clean_docstring(item._obj.__doc__)
 
-        fixtures = {fixture_name: item.funcargs[fixture_name]
-                    for fixture_name in item.fixturenames
-                    if fixture_name not in ['request',
-                                            'required_tag_names',
-                                            'pytestconfig']}
```
g-k: 👍 this is janky and blows up the result json

aws/redshift/test_redshift_security_group_does_not_allow_all_ips_access.py
ajvb: It does return a very specific error response. I'll open an issue.

g-k: 👍

I was wondering if the could detect that at the account level. Should we put EC2-Classic tests in a separate folder?

ajvb: Oh I see. From what I saw, it wasn't really exposed by the API besides in error responses, but worth further investigation.

g-k: We could also delete the test and make an issue to add it back if there's demand for testing EC2-Classic accounts. 

Dunno, skipping is good for now.

PR #71, opened 2018-02-05
Filter out INFO severity results from markdown report

ajvb:
Allows for the markdown report to be cleaner and more easily actionable.

r? @g-k


ajvb:
@g-k Oh absolutely!

Issue #67, opened 2018-02-05
Fuzzy test name and ID matchers


g-k:
refs #18 #28 

It'd be handy to have fuzzier matching on test names and IDs and (maybe markers too) in the severity and exemptions config.

We might be able to reuse the pytest substring and marker matchers. Another option would be regex support.

The tricky thing would be handling multiple matching rules and avoiding accidentally adding exemptions that mask actual failures.

Alternatively, we could improve the workflow for adding exemptions (something interactive or an option to approve and output all failures for a test in the exemption config format).

g-k:
Real world use will probably give us a better idea of how this should work.

ajvb:
@g-k yep, real world use is confirming the need for this haha. Unless you want to work on this, I'm happy to jump on it.

g-k:
@ajvb Go for it

PR #78, opened 2018-02-08
Substring matching for exemptions test id's

ajvb:
Adds the ability to prefix a test id with an asterisk
to enable substring matching.

refs #67 

r? @g-k

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,6 +232,8 @@ severity config, but requires additional columns for test ID (usually
 an AWS resource ID), exception expiration day (as YYYY-MM-DD), and
 exception reason.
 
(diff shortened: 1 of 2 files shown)


ajvb:
@g-k I don't really like this implementation that much, but it seemed the easiest without doing a bigger refactor of `exemptions.py`. Do you think it's acceptable for now?

g-k:
@ajvb I think this is fine for now.

ajvb:
@g-k 👍 Added a bit in this PR and opened an issue (https://github.com/mozilla-services/pytest-services/issues/80)

PR #81, opened 2018-02-12
Better test_ec2_security_group_in_use test

ajvb:
This improves `test_ec2_security_group_in_use` to include almost all services that can have a security group associated with it.

r? @g-k


ajvb:
@g-k I was wondering the same thing. It's just funny that most services will fit that format perfectly, with the exception of a few things under "EC2" (like sec groups, vpcs, etc).

PR #84, opened 2018-02-13
Only test secgroups attached to resources

ajvb:
This changes aws/ec2/test_ec2_security_group_opens_* tests to only fail if they
are attached to a resource. It also takes test_ec2_security_group_in_use out of
the default runs defined as make targets.

r? @g-k

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,14 +11,16 @@ all: check_venv
 	pytest
 
 awsci: check_venv
-	pytest --continue-on-collection-errors aws/ --json=results-$(AWS_PROFILE)-$(TODAY).json $(PYTEST_OPTS)
+	pytest --continue-on-collection-errors aws/ \
+		--ignore aws/ec2/test_ec2_security_group_in_use.py \
+		--json=results-$(AWS_PROFILE)-$(TODAY).json $(PYTEST_OPTS)
 
 aws-sg: check_venv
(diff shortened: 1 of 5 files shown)


ajvb:
@g-k I was wondering about that. How exactly do you see a custom marker being used?

Another alternative is to use a custom function in `aws/ec2/resources.py` that filters out secgroups where "InUse" is false:
```
def ec2_security_groups_in_use():
    return [secgroup for secgroup in ec2_security_groups_with_in_use_flag() if secgroup["InUse"]]
```

Also, on a side note, 🙇 for the caching layer. Calling `ec2_security_groups_with_in_use_flag()` a bunch of times without it would be horrid.

g-k:
I like that we skip explicitly. 

Maybe if the skip fn gets the sg:

```py3
@pytest.mark.skipif(lambda sg: not sg['InUse'])
```

or

```py3
[(pytest.mark.skip(sg) if sg['InUse'] else sg) for sg in ec2_security_groups_with_in_use_flag()]
```

in the parametrize though I'm not sure that's valid python. 

What you have is good and clear 👍  :shipit:

ajvb:
@g-k 👍 Totally. It seems like `skipif` currently only accepts a boolean, but I see what you are getting at and I'm a fan. Will revisit.

PR #96, opened 2018-02-27
meta tests

g-k:
Nice to have.

The original motivation was to test things well enough to refactor the AWS client without breaking everything at some point, but this might be helpful before then (e.g. for AWS error responses).

I have more example cached responses I can clean up and some WIP tests for the client that I could push at some point. 

Changes:

* add a way to run pytest tests (e.g. to mock and test `AWS_PROFILE` env vars) against all files in addition to doctests and track combined coverage
* add an `example_cache` so we can run tests offline and in CI

Open questions/bikeshedding: 

* Is the `meta_test` prefix weird? Should meta tests go in a separate top level folder?
* Would keeping a folder cached responses make everyone sad long term?

r? @ajvb for initial feedback

TODO:

- [ ] fix flake8 failures


ajvb:
Oh this is awesome! :+1: so far for sure.

>Is the meta_test prefix weird? Should meta tests go in a separate top level folder?

It's a little strange, but it also makes sense in this context imo.

>Would keeping a folder cached responses make everyone sad long term?

Potentially. How are you currently generating the JSON?

g-k:
Cool, I'll get the flake8 tests passing later this week.

> >    Is the meta_test prefix weird? Should meta tests go in a separate top level folder?

> It's a little strange, but it also makes sense in this context imo.

The imports would be trickier to work out, but it's certainly possible.

> >    Would keeping a folder cached responses make everyone sad long term?

> Potentially. How are you currently generating the JSON?

I took existing cached responses from stage, renamed them, pared them down to a couple objects (2 users in this case),  and mangled identifying info (arns, UIDs, usernames, etc.). 

I don't think we have to have example cached responses for all resources, but it'd be nice to have them for debugging AWS client strangeness (e.g. error responses or testing one resource with a different format without fetching all the resources for the account again). Assuming the storage format and paths stays the same it shouldn't be too painful.

g-k:
@ajvb r?

I can add more cache and test files in other PRs.

g-k:
👍 thanks!

PR #97, opened 2018-02-27
Adds new test for cross-account roles requiring MFA

ajvb:
Test that all IAM Roles that include admin policies and have cross account trust relationships require MFA.

As with the other "admin" tests, the way we are determining what role is an "admin" role is overly simplistic, but works as a first pass.

@g-k With this test and the other two `test_iam_admin_*` tests, I'd like to make sure it's known to any users that they are probably going to have false positives. Do you think adding a note about that to the README is enough?

r? @g-k


ajvb:
@g-k :+1: sounds like a plan. Added to README and docstrings

PR #99, opened 2018-03-06
test_iam_user_is_inactive with config support

### LINKED ISSUE #98: Create a configurable test around discovering inactive accounts
Create a test that will download the credential report and look for users that haven't been used in X amount of time, configurable within the config.yaml file.

ajvb:
`test_iam_user_is_inactive` will fail on users considered "inactive". What is considered inactive by default is "Hasn't used the account in over a year from when the test is run" and there is a default grace period of 1 week on new accounts (since they may not have used the account at all yet). Config supports years, months, and weeks for relative "time ago".

Fixes #98

r? @g-k

### DIFF
--- a/aws/iam/helpers.py
+++ b/aws/iam/helpers.py
@@ -1,39 +1,43 @@
 from dateutil.parser import parse
 
 
-def user_is_inactive(iam_user, grace_period, considered_inactive):
+def user_is_inactive(iam_user, no_activity_since, created_after):
     """
     Returns False if any of these are true:
-        - The user was created less than the grace period
-        - The user has used either possibly active access keys since the date
-          that is "considered_inactive"
+        - The user was created after the passed in "created_after" datetime.
+        - The user has used either potentially active access keys since the date
+          that is "no_activity_since"
         - The user has logged into the AWS console since the date that is
-          "considered_inactive"
+          "no_activity_since"
     else it will return True.
 
     >>> from datetime import datetime
-    >>> grace_period = datetime(2018, 1, 8)
-    >>> considered_inactive = datetime(2017, 1, 1)
+    >>> no_activity_since = datetime(2017, 1, 1)
+    >>> created_after = datetime(2018, 1, 8)
 
-    >>> user_is_inactive({'user_creation_time': '2018-01-10'}, grace_period, considered_inactive)
+    User considered active due to being created after the created_after datetime.
+    >>> user_is_inactive({'user_creation_time': '2018-01-10'}, created_after, no_activity_since)
     False
 
+    User considered active due to usage of access key 1 after no_activity_since
     >>> user_is_inactive({
     ...     'user_creation_time': '2016-01-10',
     ...     'access_key_1_active': 'true',
     ...     'access_key_1_last_used_date': '2017-06-01',
-    ... }, grace_period, considered_inactive)
+    ... }, no_activity_since, created_after)
     False
 
+    User considered active due to usage of access key 2 after no_activity_since
     >>> user_is_inactive({
     ...     'user_creation_time': '2010-01-10',
     ...     'access_key_1_active': 'true',
     ...     'access_key_1_last_used_date': '2014-06-01',
     ...     'access_key_2_active': 'true',
     ...     'access_key_2_last_used_date': '2017-02-01',
-    ... }, grace_period, considered_inactive)
+    ... }, no_activity_since, created_after)
     False
 
+    User considered active due to usage of password after no_activity_since
     >>> user_is_inactive({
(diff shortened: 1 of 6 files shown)


aws/iam/helpers.py
```diff
@@ -0,0 +1,85 @@
+from dateutil.parser import parse
+
+
+def user_is_inactive(iam_user, grace_period, considered_inactive):
+    """
+    Returns False if any of these are true:
+        - The user was created less than the grace period
```
g-k: If I'm understanding this correctly this is "before the account creation grace period has elapsed" or "within the account creation first use grace period."

aws/iam/helpers.py
```diff
@@ -0,0 +1,85 @@
...
+        - The user has used either possibly active access keys since the date
+          that is "considered_inactive"
+        - The user has logged into the AWS console since the date that is
+          "considered_inactive"
+    else it will return True.
+
+    >>> from datetime import datetime
+    >>> grace_period = datetime(2018, 1, 8)
```
g-k: Assuming the above comment about whether I'm understanding `grace_period` correctly then I'd expect `grace_period` to be a time interval timedelta and we'd compare something like: `parse(iam_user['user_creation_time']) + grace_period >= datetime.now()` on L71.

OK, I see this is happening the config parsing, so maybe just renaming these and `considered_inactive` to be clearer like `created_after` and `no_activity_since` would work.

aws/iam/helpers.py
```diff
@@ -0,0 +1,85 @@
...
+    ... }, grace_period, considered_inactive)
+    False
+
+    >>> user_is_inactive({
+    ...     'user_creation_time': '2010-01-10',
+    ...     'access_key_1_active': 'true',
+    ...     'access_key_1_last_used_date': '2014-06-01',
+    ...     'access_key_2_active': 'true',
+    ...     'access_key_2_last_used_date': '2017-02-01',
+    ... }, grace_period, considered_inactive)
```
g-k: These tests are great 💯

g-k: Might be useful to describe the tests like "user active due to recent use of second access key"

aws/iam/helpers.py
```diff
@@ -0,0 +1,85 @@
...
+
+    if iam_user['access_key_1_active'] == 'true' and iam_user['access_key_1_last_used_date'] != 'N/A':
+        if parse(iam_user['access_key_1_last_used_date']) > considered_inactive:
+            return False
+    if iam_user['access_key_2_active'] == 'true' and iam_user['access_key_2_last_used_date'] != 'N/A':
+        if parse(iam_user['access_key_2_last_used_date']) > considered_inactive:
+            return False
+    if iam_user['password_enabled'] == 'true' and \
+            iam_user['password_last_used'] not in ['N/A', 'no_information']:
+        if parse(iam_user['password_last_used']) > considered_inactive:
```
g-k: Maybe flatten these into one conditional and extract a small function or two to check the different parts of the conditional.

aws/iam/test_iam_user_is_inactive.py
```diff
@@ -0,0 +1,31 @@
...
+def test_iam_user_is_inactive(iam_user_row, considered_inactive, grace_period):
+    """Tests if a user is inactive. This is done by checking the last time
+    an access key was used or the user logged into the console. If the config
+    settings are not present, it defaults to considering a year ago as inactive
+    and giving a user a 1 week grace period.
+    """
+    assert not user_is_inactive(iam_user_row, grace_period, considered_inactive), \
+        "User {} hasn't been used since {} and is out of the grace period.".format(
+            iam_user_row['user'], considered_inactive.date()
+        )
```
g-k: 👍

custom_config.py
```diff
@@ -45,3 +48,25 @@ def get_whitelisted_ports_from_test_id(self, test_id):
                 return set(rule['ports'])
 
         return set([])
+
+    def considered_inactive(self):
+        considered_inactive = self._parse_user_is_inactive_relative_time('considered_inactive')
+        if considered_inactive is None:
+            return datetime.now(timezone.utc)-relativedelta(years=+1)
```
g-k: I think there's a UTC now method.

ajvb: [`datetime.utcnow()`](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) is a little different, as it's not timezone-aware. Need that since all of the times from `get_credential_report()` are timezone-aware.

aws/iam/helpers.py
ajvb: Thank you, I was struggling a bit with how to name these two vars.

PR #103, opened 2018-03-23
Initial GSuite test

ajvb:
This adds support for GSuite tests and adds the first test: `test_admin_user_is_inactive`.

r? @g-k

### DIFF
--- a/config.yaml.example
+++ b/config.yaml.example
@@ -40,3 +40,9 @@ aws:
       ports:
         - 22
         - 2222
+gsuite:
+  domain: 'example.com'
+  user_is_inactive:
+    no_activity_since:
+      years: 1
+      months: 0
--- a/conftest.py
+++ b/conftest.py
@@ -71,7 +71,7 @@ def pytest_configure(config):
 
     config.custom_config = custom_config.CustomConfig(config.getoption('--config'))
 
-    if any([x for x in config.args if 'gsuite' in x]):
+    if any(x for x in config.args if 'gsuite' in x):
         gsuite_client = GsuiteClient(domain=config.custom_config.gsuite.domain, offline=config.getoption('--offline'))
     else:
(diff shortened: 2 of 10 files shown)


config.yaml.example
```diff
@@ -40,3 +40,9 @@ aws:
       ports:
         - 22
         - 2222
+gsuite:
+  domain: 'example.com'
+  user_is_inactive:
```
g-k: I wonder if we should abstract this into a cross-provider config value in the future (since we'd probably have the same policy across providers).

conftest.py
```diff
@@ -68,6 +71,11 @@ def pytest_configure(config):
 
     config.custom_config = custom_config.CustomConfig(config.getoption('--config'))
 
+    if any([x for x in config.args if 'gsuite' in x]):
```
g-k: We should be able to drop the [] brackets since the any will consume the generator.

custom_config.py
```diff
@@ -26,29 +27,11 @@ def add_markers(self, item):
         regressions.add_regression_marker(item)
 
 
-class AWSConfig:
+class CustomConfigMixin:
```
g-k: Ah nice, disregard the first comment.

gsuite/client.py
```diff
@@ -0,0 +1,49 @@
...
+
+ADMIN_DIRECTORY_USER_READONLY = 'admin-directory-user-readonly'
+CREDENTIALS = [
+    (ADMIN_DIRECTORY_USER_READONLY, 'https://www.googleapis.com/auth/admin.directory.user.readonly'),
+]
+
+
+def get_credential_dir():
+    home_dir = os.path.expanduser('~')
+    credential_dir = os.path.join(home_dir, '.credentials')
```
g-k: this might conflict with the github org scripts filename (should be fine for now)

PR #107, opened 2018-04-03
Beginnings of implementing forensics-ready tests

ajvb:
This includes 4 new tests that are taken from prowler's
(https://github.com/toniblyx/prowler) set of "forensics-ready" tests.

These include 3 CIS Benchmark tests (2 for Cloudtrail and 1 for VPC Flow
Logs) and an additional test for logging being enabled on
ElasticSearchService Domains.

@g-k I would love your take on the current implementation of `test_cloudtrail_enabled_in_all_regions` and `test_ec2_vpc_flow_log_enabled`. I am doing a resource fetch within the test as I don't want these tests to not run if there are no cloudtrails or vpc flow logs. What do you think?

As well, tests are not passing due to the fact that `test_cloudtrail_enabled_in_all_regions` is tested on regions and there is a hardcoded list of `['us-east-1']` for when `--offline` is passed in. Will fix once I get the :+1: / :-1: from you on the above mentioned issue.

### DIFF
--- a/aws/cloudtrail/test_cloudtrail_enabled_in_all_regions.py
+++ b/aws/cloudtrail/test_cloudtrail_enabled_in_all_regions.py
@@ -1,0 +1,20 @@
+import pytest
+
+from conftest import botocore_client
+
+from aws.cloudtrail.resources import cloudtrails
+
+
+@pytest.mark.cloudtrail
+@pytest.mark.parametrize('aws_region',
+                         botocore_client.regions,
+                         ids=lambda region: region)
+def test_cloudtrail_enabled_in_all_regions(aws_region):
+    """
+    Tests that all regions have an associated cloudtrail or that there is a cloudtrail for all regions.
+    """
+    trails = cloudtrails()
+    assert any(
+        trail for trail in trails
+        if trail['HomeRegion'] == aws_region or trail['IsMultiRegionTrail']
+    )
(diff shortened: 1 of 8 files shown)


aws/cloudtrail/test_cloudtrail_enabled_in_all_regions.py
```diff
@@ -0,0 +1,20 @@
...
+
+@pytest.mark.cloudtrail
+@pytest.mark.parametrize('aws_region',
+                         botocore_client.regions,
+                         ids=lambda region: region)
+def test_cloudtrail_enabled_in_all_regions(aws_region):
+    """
+    Tests that all regions have an associated cloudtrail or that there is a cloudtrail for all regions.
+    """
+    trails = cloudtrails()
```
g-k: Yeah, as you suspected this and the EC2 test are kind of weird to me.

I guess an unwritten goal is to avoid IO in the tests, so they can be pure functions on data. Though this isn't set in stone.

Is the fetch so we never use cached data for these endpoints? If so, I think there's still a use_cache kwd arg in to the client we could try or we could build a top level non-caching client to call instead.

ajvb: I took another look at this today and realized I was just being dumb. I wanted to make sure that these tests always ran, even when there were 0 cloudtrails or 0 VPC flow logs, but in testing them along side regions / vpcs this issue is resolved. Just pushed up a fix.

ajvb:
@g-k r? - should be ready for a proper review now :)

aws/ec2/test_ec2_vpc_flow_log_enabled.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+
+from aws.ec2.resources import (
+    ec2_flow_logs,
+    ec2_vpcs
+)
+
+
+@pytest.fixture
```
g-k: 👍

PR #149, opened 2018-08-22
Pagerduty bug 1471730

g-k:
Changes:

* replace pagerduty example test with two tests using json files
* drop pygerduty dep

NB: this doesn't check for misconfigured or extra pagerduty users (@ameihm0912 should we be checking that?). It just checks for users with IAM or bastion access. 

r? @ajvb


pagerduty/test_iam_users_have_pd_remote_monitoring_configured.py
```diff
@@ -0,0 +1,22 @@
...
+from pagerduty.helpers import alternate_usernames
+from pagerduty.resources import (
+    users_with_remote_access_escalation_policy_and_extension_configured
+)
+
+
+@pytest.mark.pagerduty
+@pytest.mark.parametrize("iam_user", list(get_all_users_that_can_access_aws_account()))
+def test_bastion_users_have_pd_remote_monitoring_configured(
+    iam_user, users_with_remote_access_escalation_policy_and_extension_configured
```
g-k: Some `iam_user`s are returning as email addresses, so we'll probably need a bunch of alternates or want to strip trailing `@mozilla.{com,org}`

README.md
```diff
@@ -438,6 +442,77 @@ gsuite:
...
+  users_with_remote_access_monitoring: 'pd_users.json'
+  bastion_users: 'hierahash/*hierahash.json'
+  alternate_usernames: 'alternate_usernames.json'
+```
+
+Where `users_with_remote_access_monitoring` and `bastion_users` are
+globs for multiple files relative to the current working directory and
+`alternate_usernames` is the path to a single file.
+
+The files have examples formats as follows:
```
g-k: currently planning to generate these outside pytest-services.

pagerduty/test_iam_users_have_pd_remote_monitoring_configured.py
```diff
@@ -0,0 +1,22 @@
...
+from aws.iam.resources import get_all_users_that_can_access_aws_account
+from pagerduty.helpers import alternate_usernames
+from pagerduty.resources import (
+    users_with_remote_access_escalation_policy_and_extension_configured
+)
+
+
+@pytest.mark.pagerduty
+@pytest.mark.parametrize("iam_user", list(get_all_users_that_can_access_aws_account()))
+def test_bastion_users_have_pd_remote_monitoring_configured(
```
ajvb: I believe this should be `test_iam_users...` rather than `test_bastion_users...`

pagerduty/helpers.py
```diff
@@ -0,0 +1,21 @@
...
+
+import functools
+import json
+from pathlib import Path
+
+import pytest
+
+
+@functools.lru_cache()
+def _altnames():
```
ajvb: nit: should `_altnames()` and `alternate_usernames()` be in `pagerduty/resources.py`? Seem more like a resource than a helper but ¯\_(ツ)_/¯

README.md
```diff
@@ -438,6 +442,77 @@ gsuite:
...
+```json
+{
+    "chobbes": {
+        "groups": [
+            ""
+        ],
+        "root_ssh": true
+	},
+	"movedon": {
+        "ensure": "absent"
```
ajvb: nit: is this moving a closing `}`?

pagerduty/helpers.py
g-k: It is, but I was considering it a private resource with the leading `_`.

I'll make it public and move it to resources.

pagerduty/helpers.py
```diff
@@ -0,0 +1,21 @@
...
+    # TODO: using pytest.config directly isn't recommended
```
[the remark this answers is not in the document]
g-k: I think it's fine since we're not using pytest in a normal way anyway, so I'll just drop the TODO.

Is `pytestconfig` available outside fixtures?

pagerduty/test_iam_users_have_pd_remote_monitoring_configured.py
g-k: Yep, good catch!

README.md
g-k: Yeah, the whitespace is off here. I'll clean that up.

pagerduty/helpers.py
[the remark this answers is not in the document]
ajvb: Ok totally. Oh, good point no I don't think it is.

README.md
g-k: mixed tabs and spaces :scream:

PR #154, opened 2018-10-01
Gsuite: support new directory test

ajvb:
* Refactor how credentials are created, one set of credentials rather
than one per scope.
* Add support for test_groups_have_enough_owners
* Update google_api_client lib

### DIFF
--- a/bin/auth/setup_gsuite.py
+++ b/bin/auth/setup_gsuite.py
@@ -1,7 +1,8 @@
 import os
 
 from gsuite.client import (
-    CREDENTIALS,
+    CREDS_NAME,
+    SCOPES,
     get_credentials,
     get_credential_dir,
     get_credential_path,
@@ -30,8 +31,7 @@ def get_or_create_credentials(credential_name, scopes):
 
 
 def main():
-    for creds in CREDENTIALS:
-        get_or_create_credentials(creds[0], creds[1])
+    get_or_create_credentials(CREDS_NAME, SCOPES)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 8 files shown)


bin/auth/setup_gsuite.py
```diff
@@ -30,8 +31,7 @@ def get_or_create_credentials(credential_name, scopes):
 
 
 def main():
-    for creds in CREDENTIALS:
-        get_or_create_credentials(creds[0], creds[1])
+    get_or_create_credentials(CREDS_NAME, SCOPES)
```
g-k: :+1:

gsuite/admin/helpers.py
```diff
@@ -6,3 +6,21 @@ def user_is_inactive(user, no_activity_since):
...
+
+
+def is_owner_of_group(member):
+    """
+    Check whether a member of a group is an owner with a status of 'ACTIVE'.
+    """
+    return (
+        member["type"] == "USER"
+        and member["role"] == "OWNER"
+        and member["status"] == "ACTIVE"
```
g-k: Not sure how many of these strings there are, but we might want to throw them in [an enum](https://docs.python.org/3/library/enum.html) at some point.

gsuite/client.py
```diff
@@ -48,4 +47,55 @@ def list_users(self):
...
+            while True:
+                pageToken = resp["nextPageToken"]
+                resp = (
+                    self.directory_client.members()
+                    .list(groupKey=group, pageToken=pageToken)
+                    .execute()
+                )
+                members += resp["members"]
+                if not resp.get("nextPageToken", ""):
+                    break
```
g-k: Surprised the goog api client doesn't do this. Should make an issue for de-duping this pagination logic.

ajvb: My bad. The documentation for this library is super scattered and I missed that it does have pagination helpers. Pushed up a fix

g-k: :+1: no worries

PR #159, opened 2018-11-12
Upgrade packages

ajvb:
Includes an update to our use of pytest's Cache


ajvb:
Fixes #146 #147 #150 #153 #157

PR #239, opened 2019-03-20
Update required ami test logic

jvehent:
This works but the list of owner accounts should really come from a config file. I didn't yet take the time to figure that part out.


aws/ec2/resources.py
```diff
@@ -117,7 +118,8 @@ def ec2_security_groups_with_in_use_flag():
 
     # These resources have two types of security groups, therefore
     # the Vpc ones are namespaced under "VpcSecurityGroups"
-    vpc_namespaced_resources = sum([rds_db_instances(), redshift_clusters()], [])
+    vpc_namespaced_resources = sum(
+        [rds_db_instances(), redshift_clusters()], [])
```
jvehent: this and the above are automated formatting from vscode. should I remove them?

aws/ec2/resources.py
```diff
@@ -135,3 +137,14 @@ def ec2_security_groups_with_in_use_flag():
             sec_group["InUse"] = False
 
     return sec_groups
+
+
+def ec2_images_owned_by():
+    "Returns a list of EC2 images owned by a list of provided accounts"
+    return (
+        botocore_client.get("ec2", "describe_images", [],
+                            {"Filters": [{"Name": "owner-id", "Values": ["361527076523", "351644144250", "178550527431", "342354050843", "927034868273"]}]})
```
jvehent: we need to get that list from a config file somehow. @ajvb is that something you could quickly add to this patch?

Also, I tried adding a long list of account IDs here, and it seems to make pytest fail to create cache files because it tried to put all the IDs in the filename.

aws/ec2/resources.py
ajvb: Yeah, we use [black](https://github.com/ambv/black) for formatting, with https://github.com/mozilla-services/pytest-services/blob/master/Makefile#L51-L52 checking in CI.

aws/ec2/resources.py
ajvb: Yeah absolutely. Do we actually want more account IDs than the one we are scanning against?

aws/ec2/test_ec2_instance_running_required_amis.py
```diff
@@ -1,25 +1,39 @@
...
-def required_amis(pytestconfig):
-    return frozenset(pytestconfig.custom_config.aws.required_amis)
+def owned_amis():
+    return ec2_images_owned_by()
 
 
 @pytest.mark.ec2
 @pytest.mark.parametrize("ec2_instance", ec2_instances(), ids=ec2_instance_test_id)
-def test_ec2_instance_running_required_amis(ec2_instance, required_amis):
+def test_ec2_instance_on_acceptable_ami(ec2_instance, owned_amis,):
```
ajvb: Should change the file name to match the test name.

aws/ec2/test_ec2_instance_running_required_amis.py
```diff
@@ -1,25 +1,39 @@
...
-    assert (
-        ec2_instance["ImageId"] in required_amis
-    ), "EC2 Instance {0[InstanceId]} is running unexpected AMI: {0[ImageId]}".format(
-        ec2_instance
-    )
+    for tag in ec2_instance["Tags"]:
+        if tag["Key"] == "Name":
+            instanceName = tag["Value"]
+
+    nightyDaysAgo = datetime.now() - timedelta(days=180)
```
ajvb: We probably will want to make this configurable with 180 days as the default. I can make that change when I add the account ids for the amis request.

PR #240, opened 2019-03-21
Initial start on FAQ

hwine:
A start at addressing two problems:
  - Providing an FAQ for others to use
  - Documenting the answers to questions I've found

This PR is likely to stay in _draft_ mode for quite a while.


hwine:
deleting from my fork, will reopen from branch in main repo

Issue #116, opened 2018-05-15
document adding a new service


g-k:

hwine:
I'll set this up as a normal rst -> sphinx -> readthedocs doc set unless someone hollers _REAL_ loud.

g-k:
> I'll set this up as a normal rst -> sphinx -> readthedocs doc set unless someone hollers REAL loud.

👍  yeah, we've just been throwing everything in the readme, but that'd be best since this is a python project.

hwine:
> since this is a python project.

pshaw -- Firefox is [documented](https://github.com/mozilla/gecko-dev/search?q=filename%3A*.rst) in sphinx! 😉

ajvb:
@hwine looks good. May want to call out that you are not creating a new `conftest.py`, but editing the one at the top level of the repo.

ajvb:
I believe this is done?

hwine:
Yes -- this issue was satisfied by #318, and docs are at https://mozilla.github.io/frost/_build/html/index.html

PR #241, opened 2019-03-22
Initial start on FAQ

hwine:
A start at addressing several problems:
  - Continuation of issue #116 work (existing [docs][docs] were not sufficient for me to re pick up the migration work)
  - Providing an FAQ for others to use
  - Documenting the answers to questions I've found

This PR is likely to stay in _draft_ mode for quite a while. This may yet split into individual overview, service addition, and operation FAQs

[docs]: https://github.com/mozilla-services/pytest-services/blob/docs/docs/NewServices.rst


hwine:
I'm unsure why it looks like y'all committed on this branch, let me clean that up first :/

PR #246, opened 2019-04-29
Initial pass on refreshing GCP tests

ajvb:
Adds:
  * `test_firewall_opens_all_ports_to_any`
  * `test_firewall_opens_any_ports_to_all`
  * `test_admin_service_accounts`
  * `test_sql_instance_private_ip_required`
  * Initial GCP docs in README

Improve:
  * Fix error within conftest when no args are provided to pytest
  * Add error message to `test_sql_instance_ssl_required`

### DIFF
--- a/gcp/compute/helpers.py
+++ b/gcp/compute/helpers.py
@@ -1,3 +1,36 @@
+def does_firewall_open_all_ports_to_any(firewall):
+    """
+    Returns True if firewall has a rule to open all ports to any source. Excludes ICMP.
+
+    >>> does_firewall_open_all_ports_to_any({})
+    False
+    >>> does_firewall_open_all_ports_to_any({'sourceRanges': ['1.1.1.1/1'], 'allowed': [{'ports': ['1', '2', '300']}]})
+    False
+    >>> does_firewall_open_all_ports_to_any({'sourceRanges': ['1.1.1.1/1'], 'allowed': [{'ports': ['0-65535']}]})
+    True
+    >>> does_firewall_open_all_ports_to_any({'sourceRanges': ['0.0.0.0/0'], 'allowed': [{'ports': ['0-65535']}]})
+    True
+    >>> does_firewall_open_all_ports_to_any({'sourceRanges': ['10.0.0.5/32'], 'allowed': [{'ports': ['0-65535']}]})
+    True
+    """
+    if does_firewall_open_all_ports_to_all(firewall):
+        return True
+
+    if firewall.get("sourceRanges") is None:
+        return False
+
+    for rule in firewall.get("allowed"):
+        if rule.get("IPProtocol", "") == "icmp":
+            continue
+        if not rule.get("ports"):
+            return True
+        for port_rule in rule.get("ports"):
+            if port_rule == "0-65535":
+                return True
+
+    return False
+
+
 def does_firewall_open_all_ports_to_all(firewall):
     """
     Returns True if firewall has a rule to open all ports to all. Excludes ICMP.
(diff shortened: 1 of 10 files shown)


gcp/compute/helpers.py
```diff
@@ -1,3 +1,36 @@
+def does_firewall_open_all_ports_to_any(firewall):
+    """
+    Returns True if firewall has a rule to open all ports to any source. Excludes ICMP.
+
+    >>> does_firewall_open_all_ports_to_any({})
+    False
```
g-k: Does GCP default to closed? 

I think AWS default to all outbound open.

gcp/compute/helpers.py
```diff
@@ -1,3 +1,36 @@
...
+    if firewall.get("sourceRanges") is None:
+        return False
+
+    for rule in firewall.get("allowed"):
+        if rule.get("IPProtocol", "") == "icmp":
+            continue
+        if not rule.get("ports"):
+            return True
+        for port_rule in rule.get("ports"):
+            if port_rule == "0-65535":
```
g-k: does GCP normalize this to one range? e.g. could someone write `0-1,1-65535` and we wouldn't pick it up

gcp/compute/helpers.py
```diff
@@ -1,3 +1,36 @@
...
+    >>> does_firewall_open_all_ports_to_any({})
+    False
+    >>> does_firewall_open_all_ports_to_any({'sourceRanges': ['1.1.1.1/1'], 'allowed': [{'ports': ['1', '2', '300']}]})
+    False
+    >>> does_firewall_open_all_ports_to_any({'sourceRanges': ['1.1.1.1/1'], 'allowed': [{'ports': ['0-65535']}]})
+    True
+    >>> does_firewall_open_all_ports_to_any({'sourceRanges': ['0.0.0.0/0'], 'allowed': [{'ports': ['0-65535']}]})
+    True
+    >>> does_firewall_open_all_ports_to_any({'sourceRanges': ['10.0.0.5/32'], 'allowed': [{'ports': ['0-65535']}]})
+    True
```
g-k: Do we need to worry about IPv6?

gcp/compute/test_firewall_opens_all_ports_to_any.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+@pytest.mark.gcp_compute
+@pytest.mark.parametrize("firewall", in_use_firewalls(), ids=firewall_id)
+def test_firewall_opens_all_ports_to_any(firewall):
+    """
+    This test confirms that there are no firewall
+    rules that allow ingress access to all ports from anywhere.
+
+    A part of CIS 3.1
```
g-k: nit: should we tag CIS-related tests?

gcp/compute/helpers.py
ajvb: @g-k You mean if no rules are present?

gcp/compute/helpers.py
ajvb: >does GCP normalize this to one range? e.g. could someone write 0-1,1-65535 and we wouldn't pick it up

GCP does not, but the default rules / docs point towards using `0-65535`. I think this is ok for now though. If any ports except 80 or 443 are opened to the public internet, `test_firewall_opens_any_ports_to_all` will catch it. We will miss big ranges allowed internally, but atm I think that's ok (and it's also true within our AWS tests)

gcp/compute/test_firewall_opens_all_ports_to_any.py
ajvb: I think it's a nice reference, especially for newer (for us) services that have CIS Benchmarks.

gcp/compute/helpers.py
ajvb: Not for these tests, but thanks for bringing this up! Needed to confirm that. IPv6 is not supported within VPC networks, but is supported for load balancers, so if/when tests are added for those IPv6 will need to be worried about.

gcp/compute/helpers.py
ajvb: From https://cloud.google.com/vpc/docs/firewalls#default_firewall_rules - the networks have default deny all.

PR #259, opened 2019-05-30
If a firewall is disabled, ignore it

ajvb:


gcp/compute/resources.py
```diff
@@ -27,6 +27,8 @@ def networks_with_instances():
 
 def in_use_firewalls():
     for firewall in firewalls():
+        if firewall["disabled"] == True:
+            continue
         for network in networks_with_instances():
```
hwine: aiui from slack, we're adding this for a "one time" use case: needed for duration of a DB Migration.

As I read this snippet we're trading a one time use case for the possibility of false negatives if production firewall rules accidentally get disabled. That seems a case of "the cure is worse than the disease" ☹️ 

Are other options not available? Like (from my imagination - I don't know feasibility):
- Have the unused firewall "clause" deleted from this one place.
- Add this one case "violation" to the exception list to quiet the false positive.

I don't understand enough about the intricacies to give an "r-", but I'd love to see these questions answered.

g-k: Maybe we should track disabled firewalls separately?

jvehent: Maybe we shouldn't allow for firewalls to be disabled? At any rate, I'm curious what cloudops thinks of the pattern, I'll ask at the next team mtg.

ajvb: Not allowing disabled firewall rules would be a good idea and I can easily write another test to detect those. This approach of not allowing disabled firewall rules works for GCP since there are so few projects, but I'd like to keep this pattern in here as it is what we do in AWS and would prefer if we didn't diverge.

>As I read this snippet we're trading a one time use case for the possibility of false negatives if production firewall rules accidentally get disabled.

I'm not sure I understand this. We are testing only for overly permissive firewall rules, so if it's accidentally disabled that's fine. Ops will catch that. If a firewall rule was created in the past, but is now disabled, it's only a false negative for the time in-between one run and the next if it was re-enabled. So, it's not a false negative if it stays disabled.

This pattern was more important in AWS as there were at times 100's of security groups that were overly permissive that were not attached to any instances. Instead of flagging them, we only test "in use" or "active" resources (this includes filtering out EC2 instances that are shutdown, user accounts from MFA checks if they can't log into the console, etc).

As well, this pattern of only testing "in use" or "active" resources has the benefit of nearly all of the failed tests are true positives that you want to get fixed. I believe this is one of the biggest benefits of this tool over something like ScoutSuite. Lastly, if something gets re-activated, we flag it within 24 hours as the tests run nightly.

PR #265, opened 2019-06-27
report on GitHub compliance via pytest-services 

hwine:
Part of #1351 

**Status**: working, _but not production ready_

**To Do:**
- [ ] hook up, with other repos, to a Jenkins staging build
- Repeat until done:
    - [ ] resolve issues
    - [ ] clean up code
    - [ ] make sure documentation updated
    - [ ] get code review
- [ ] merge to master
- [ ] deploy to production

**Related:**
- [PR on foxsec](https://github.com/mozilla-services/foxsec/pull/1352)
- [Changes to foxsec-tools](https://github.com/mozilla-services/foxsec-tools/compare/master...hwine:hwine/github_1) *nothing yet, so no PR*


hwine:
@ajvb I'm ready for any general comments you might have -- especially about "what goes in which file".

It's not quite read for review-before-merge, but if you have some time, I'd appreciate it.

ajvb:
@hwine sure thing, but I'd also recommend merging master in soonish, as some files like `.travis.yml` are pretty out of date.

conftest.py
```diff
@@ -26,6 +28,13 @@ def pytest_addoption(parser):
                      type=str,
                      help='Set GCP project to test. Required for GCP tests.')
 
+    # While only used for Heroku at the moment, GitHub tests are soon to be
+    # added, which will also need an "organization" option. Current plan is to
+    # reuse this one.
+    parser.addoption('--organization',
```
ajvb: let's call this `--heroku-organization`

github/resources.py
```diff
@@ -0,0 +1,132 @@
...
+    return org, date
+
+def get_data_for_org(org=None, date=None):
+    org, date = finalize_defaults(org, date)
+    srcname = f"{date}-{org}.db.obj.json"
+    data = get_data_from_file(srcname)
+    return data
+
+
+@functools.lru_cache()
```
ajvb: I don't think you will want to cache this, since it's reading from a file?

ajvb: Same goes for nearly all functions in this file.

github/resources.py
```diff
@@ -0,0 +1,132 @@
...
+import io
+import json
+import os
+
+import pyjq
+
+
+results_dir = os.environ["RESULTS_DIR"]
+organization = os.environ["Organization"]
+today = os.environ["TODAY"]
```
ajvb: Lets move these env vars to cli flags that live in `conftest.py`. Something like `--github-organization` and `--github-data-directory`

github/test_branch_protection.py
```diff
@@ -0,0 +1,83 @@
...
+        """
+Consistency check for metadata. If we're examining an org, there should
+be repos we care about in it. (Warning)
+"""
+)
+@pytest.mark.parametrize(
+        "repo_list",
+        [(resources.repos_of_interest_for_org()),],
+)
+def test_repos(repo_list):
```
ajvb: Our standard as been one test per file with the file name matching the test name.

github/test_branch_protection.py
```diff
@@ -0,0 +1,83 @@
...
+        select(.url == "/repos/{org}/{repo}/branches/{branch}") # just this branch
+        """
+    branch_data = pyjq.all(jq_branch_record, org_json)
+    assert len(branch_data) == 1
+    assert branch_data[0]["body"]["protected"]
+
+
+
+@pytest.mark.github_org
+@pytest.mark.xfail(reason="Archived repo in metadata active list")
```
ajvb: As discussed, should probably remove this `xfail`'s

github/test_download_s3.py
```diff
@@ -0,0 +1,72 @@
...
+- can we still retrieve the file
+- can we still parse the file
+- is the data of a size and format we expect
+"""
+)
+@pytest.mark.parametrize(
+        "org,date",
+        [(r.cur_org(), r.report_date()),],
+)
+def test_valid_org_file(org, date):
```
ajvb: Same as above, one test per file.

github/test_download_s3.py
```diff
@@ -0,0 +1,72 @@
...
+    # all the checks to see if we have sane input before proceeding
+    org_data = r.get_data_for_org(org, date)
```
ajvb: Our standard approach has been for the function in parameterize to handle all fetching of data.

github/test_download_s3.py
```diff
@@ -0,0 +1,72 @@
...
+import resources as r
+import pyjq
+import pytest
+
+
+
+@pytest.mark.github_org
+@pytest.mark.rationale(
+        """
+Sanity check for early development
```
ajvb: Is this entire file to be removed?

heroku/test_heroku_2fa.py
```diff
@@ -0,0 +1,51 @@
...
+
+    Returns: a list of member dictionaries
+    """
+    just_key = raw_json_data.extract_key(heroku_client.data_set_names.USER, [])
+    members = list(*just_key.results)
+
+    return members
+
+
+@pytest.mark.parametrize("a_user", list(users(raw_json_data())))
```
ajvb: do you need the `list()` call here?

requirements.txt
```diff
@@ -10,3 +10,10 @@ pytest-json==0.4.0
 pytest-metadata==1.5.1
 pytest==3.3.2
 python-dateutil==2.6.1
+git+https://github.com/hwine/python-herokuadmintools.git#egg=herokuadmintools
+wheel==0.33.4
+# We want pyjq==2.3.1, but it can't be built in the container (we don't
+# want to load all the compilers, etc). So assume the wheel is prebuilt
+# in the RESULTS_DIR, which requires pip >= 10 per
+#   https://pip.pypa.io/en/stable/reference/pip_install/#using-environment-variables
+${RESULTS_DIR}/pyjq-2.3.1-cp36-cp36m-linux_x86_64.whl
```
ajvb: I'm not a fan of this, makes it really hard for people to use outside of our team.

I'd say we load the compilers in our container

github/resources.py
hwine: @ajvb What's our pattern here? env vars to pass into container and then restate as explicit options on command line?

Are option defaults allowed? (i.e. take from env if not supplied on command line?)

github/test_branch_protection.py
hwine: Curious as to reason. I can see one concept per file, but I don't see the value an explosion of files provides.

github/test_branch_protection.py
hwine: Yep -- they're out in my current work.

github/test_download_s3.py
hwine: Let's discuss this -- that doesn't work well for arrays, and this relates to how I've tweaked the handling of metadata.

Also, this test is confirm that the infrastructure has provided the correct file for the run. (The wrapper downloads the org data before a run. If the two get out of sync -- problems.)

github/test_download_s3.py
hwine: I'll rephrase the comment -- originally yes, but there is so much scaffolding outside the pytest for file delivery and here I consume _our_ JSON files, which are not as strongly versioned as AWS/GCP, so I'm going to leave them in.

heroku/test_heroku_2fa.py
hwine: Yes, otherwise it's an OrderedDict object (or something like that)

requirements.txt
hwine: Let's discuss. I don't believe in leaving compilers in production images. We could do it with multi-stage docker images, but that's harder to maintain.

And, based on my experience, this code is really hard to reuse without documentation, so I'm not sure we can achieve purity here.

conftest.py
hwine: Why not use the generic? Both GitHub & Heroku need that as a prameter. Are we envisioning a future where one pytest invocation tests everything?

github/resources.py
hwine: The file contents don't change during a run. Actually, I've removed this cache, and cache only after the expensive operation of converting the file contents to JSON. The file access routine is only called once per file.

docs/Adding_a_Service.rst
```diff
@@ -0,0 +1,100 @@
...
+|file                                                      | action | description                                                |
++==========================================================+========+============================================================+
+| pytest-services/{NEW_SERVICE}                            | create | directory to hold the service specific files               |
++----------------------------------------------------------+--------+------------------------------------------------------------+
+| pytest-services/{NEW_SERVICE}/test_*.py                  | write  | your test cases                                            |
++----------------------------------------------------------+--------+------------------------------------------------------------+
+| pytest-services/{NEW_SERVICE}/resources.py               | write  | Code to fetch & format the data you need to test           |
++----------------------------------------------------------+--------+------------------------------------------------------------+
+| pytest-services/{NEW_SERVICE}/helpers.py                 | write  | Code to simplify writing tests                             |
```
ajvb: May not need to be mentioned here, but this is also done so that we can have doctests for the bulk of the test code logic. "Test the tests!" :tm:  @g-k

docs/Adding_a_Service.rst
```diff
@@ -0,0 +1,100 @@
...
+base docker container except as noted.
+
++----------------------------------------------------------+--------+------------------------------------------------------------+
+|file                                                      | action | description                                                |
++==========================================================+========+============================================================+
+|foxsec/tools/pytest-services/{NEW_SERVICE}                | create | directory to hold the service specific run files           |
++----------------------------------------------------------+--------+------------------------------------------------------------+
+|foxsec/tools/pytest-services/{NEW_SERVICE}/run.sh         | write  | code to invoke pytest for your service                     |
++----------------------------------------------------------+--------+------------------------------------------------------------+
+|foxsec/tools/pytest-services/{NEW_SERVICE}/config.yaml    | write  | expected failure information                               |
```
ajvb: I think that "expected failure information" as a description is disregarding a lot of options that live in the config file. Might be better to just describe as "pytest-services config file" with a link to the example config, or similar.

docs/Adding_a_Service.rst
```diff
@@ -0,0 +1,100 @@
...
+Recap of PyTest Services
+------------------------
+
+The expected workflow for a pytest service is:
+
+-   ``pytest`` runs, and produces a results*.json file:
+      - fetch any needed data
+      - test that data
+      - for failures, see if the failure is expected
+-   ``generate_service_report`` runs, and transforms the json file into a compact
```
ajvb: Might be worth linking to this script?

docs/Adding_a_Service.rst
hwine: Yes, that should be mentioned somewhere here. I have notes on which files typically get them to add in. That was part of what helped me see the differences.

docs/Adding_a_Service.rst
hwine: _Interesting_ -- I thought it was just "exemptions" plus whatever random data my tests needed. 😄 

I see more documentation forthcoming! That's a good thing!

github/test_download_s3.py
```diff
@@ -0,0 +1,72 @@
+"""
+make sure we can download a file on demand
+"""
+import resources
+import pyjq
+import pytest
+
+
```
ajvb: as discussed, might want to include a comment explaining that this is testing pytest-services stuff rather than github

conftest.py
ajvb: As discussed, separate github and heroku args for consistency

github/resources.py
ajvb: `env vars to pass into container and then restate as explicit options on command line?` - yes!

github/test_branch_protection.py
ajvb: discussed, one test per file

requirements.txt
ajvb: as discussed, should just install pyjq

PR #268, opened 2019-07-01
Add new gcp tests, update service_report_generator for gcp

ajvb:
Adds two new tests:
  * test_only_allowed_org_accounts - Confirms that only approved domains/gsuites have access to projects (with a list of role exceptions, e.g. log viewer)
  * test_gke_version_up_to_date - Confirms GKE Clusters master and node versions are in the list of valid versions.

Also adds basic support for GCP CSV reports via `service_report_generator.py`

### DIFF
--- a/gcp/iam/test_only_allowed_org_accounts.py
+++ b/gcp/iam/test_only_allowed_org_accounts.py
@@ -28,4 +28,6 @@ def test_only_allowed_org_accounts(iam_binding, allowed_org_domains):
             if not member.startswith("serviceAccount"):
                 assert (
                     member.split("@")[-1] in allowed_org_domains
-                ), "{} was found and is not in the allowed_org_domains".format(member)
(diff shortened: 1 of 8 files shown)


gcp/iam/test_only_allowed_org_accounts.py
```diff
@@ -0,0 +1,33 @@
...
+
+from gcp.iam.resources import project_iam_bindings
+
+
+@pytest.fixture
+def allowed_org_domains(pytestconfig):
+    return pytestconfig.custom_config.gcp.allowed_org_domains
+
+
+EXCLUDED_ROLES = ["roles/logging.viewer"]
```
g-k: might want to add this to the config at some point

ajvb: agreed, will most likely be following up with an addition or two to this list, will make the change then.

PR #285, opened 2019-08-13
Add test for publicly available ebs snapshots

ajvb:
Adds a new test for publicly available EBS snapshots.

Docs/Reference:
https://techcrunch.com/2019/08/09/aws-ebs-cloud-backups-leak/
https://www.cloudconformity.com/conformity-rules/EBS/public-snapshots.html#

### DIFF
--- a/aws/ec2/resources.py
+++ b/aws/ec2/resources.py
@@ -52,6 +52,36 @@ def ec2_ebs_volumes():
     )
 
 
+def ec2_ebs_snapshots():
+    "http://botocore.readthedocs.io/en/latest/reference/services/ec2.html#EC2.Client.describe_snapshots"
+    return (
+        botocore_client.get("ec2", "describe_snapshots", [], {"OwnerIds": ["self"]})
+        .extract_key("Snapshots")
+        .flatten()
+        .values()
+    )
+
+
+def ec2_ebs_snapshots_create_permission():
+    "https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_snapshot_attribute"
+    return sum(
+        [
+            botocore_client.get(
+                service_name="ec2",
+                method_name="describe_snapshot_attribute",
+                call_args=[],
+                call_kwargs={
+                    "Attribute": "createVolumePermission",
+                    "SnapshotId": snapshot["SnapshotId"],
+                },
+                regions=[snapshot["__pytest_meta"]["region"]],
+            ).values()
+            for snapshot in ec2_ebs_snapshots()
+        ],
+        [],
+    )
+
+
 def ec2_flow_logs():
     "https://botocore.readthedocs.io/en/latest/reference/services/ec2.html#EC2.Client.describe_flow_logs"
     return (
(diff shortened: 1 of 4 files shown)


aws/ec2/resources.py
```diff
@@ -52,6 +52,36 @@ def ec2_ebs_volumes():
     )
 
 
+def ec2_ebs_snapshots():
+    "http://botocore.readthedocs.io/en/latest/reference/services/ec2.html#EC2.Client.describe_snapshots"
+    return (
+        botocore_client.get("ec2", "describe_snapshots", [], {"OwnerIds": ["self"]})
```
hwine: Why are we restricting to "self" here? How are we guaranteed to iterate over all possible ownerid's? The docs suggest "self" filters out a lot of cases. The amazon owned ones we would want to exclude, but how do we know we aren't excluding volumes we care about?

It's not clear to me if ownerids are distinct from account ids, so I'm assuming:
- account ids are a subset of possible owner ids,
- the entire pytest aws suite is iterated over by account ids

While the code is clear, understanding what is the scope of testing is not. I'm not sure if that's solvable, or comprehension requires a deep mental model of all AWS objects.

aws/ec2/test_ec2_ebs_snapshots_are_private.py
```diff
@@ -0,0 +1,16 @@
...
+@pytest.mark.ec2
+@pytest.mark.parametrize(
+    "ec2_ebs_snapshot",
+    ec2_ebs_snapshots_create_permission(),
+    ids=lambda ebs: ebs["SnapshotId"],
+)
+def test_ec2_ebs_snapshot_are_private(ec2_ebs_snapshot):
+    assert not is_ebs_snapshot_public(
+        ec2_ebs_snapshot
+    ), "Snapshot {} is publicly accessible.".format(ec2_ebs_snapshot["SnapshotId"])
```
hwine: Do we need the ownerid for processing the report?

aws/ec2/resources.py
ajvb: owner id == AWS account ID. In this test, we only care about EBS snapshots from "our" EBS volumes. "Our" in this case meaning the account that is being tested.

aws/ec2/test_ec2_ebs_snapshots_are_private.py
ajvb: No, snapshot id should be enough. This is also consistent with most of our other AWS/GCP test output.

ajvb:
>I'm assuming it runs okay locally, but any false positives? Or false negatives?

No, but this just tests for the "bad" case of ebs snapshots being publicly available. You could for example accidentally share an EBS snapshot with a bad AWS account, but that is a different case then what I'd like to test for here.

aws/ec2/test_ec2_ebs_snapshots_are_private.py
hwine: Just curious - how do you know where to file the issue? Wouldn't you have to look that up?

ajvb: I filed #286 as a follow-up to this, but essentially for our use of pytest-services, we are only ever running the test suite against a specific account. This is true locally and in our nightly scans. If you look at the Bugzilla bugs (can slack you one), it specifies the aws account name and region in the finding.

hwine: Got it - I forgot about the args being in the formatted output.

PR #299, opened 2020-06-24
Test for ELB with 0 instances attached

ajvb:
Checks whether an ELB (v1) has any instances attached and fails if there are none. Similar to our "security group in use" test.


ajvb:
lol, git config busted. Recreating this PR.

PR #300, opened 2020-06-24
Test for ELB with 0 instances attached

ajvb:
Checks whether an ELB (v1) has any instances attached and fails if there are none. Similar to our "security group in use" test.


ajvb:
@g-k This is only failing because black needs to be upgraded (https://github.com/mozilla/frost/pull/301). Going to merge so that I can finish preparing for the demo tomorrow, but will happily back it all out tomorrow morning if desired/needed.

PR #305, opened 2020-07-17
Makefile should report non prod python versions

### LINKED ISSUE #304: Makefile should warn about mismatched python versions
Currently, we run Python 3.6 in production. However, that hasn't been captured in the project. It has been easy to use a different python version.

This really only makes a difference for dev & production -- CI doesn't matter as long as it supports the production version.

hwine:
Fixes #304

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ metatest:
 # that's likely to break CI.
 venv:
 	python3 -m venv venv
-	./venv/bin/python -V | grep $(PRODUCTION_PYTHON) || true; $(PYTHON_VER_ERROR)
+	./venv/bin/python -V | grep $(PRODUCTION_PYTHON) || true; $(PYTHON_VER_WARNING)
 
 .PHONY:
 	all \
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,12 @@ AWS_PROFILE := default
 
 PYTEST_OPTS := ''
 
+PRODUCTION_PYTHON := 3.6
+PYTHON_VER_WARNING = $(warning Warning! Frost production uses Python $(PRODUCTION_PYTHON), \
+		      you're running $(shell python -V))
+PYTHON_VER_ERROR = $(error Frost production uses Python $(PRODUCTION_PYTHON), \
+		      you're running $(shell python -V))
+
 all: check_venv
 	pytest
 
@@ -18,6 +24,8 @@ awsci: check_venv
 check_venv:
 ifeq ($(VIRTUAL_ENV),)
 	$(error "Run frost from a virtualenv (try 'make install && source venv/bin/activate')")
+else
+	python -V | grep $(PRODUCTION_PYTHON) || true ; $(PYTHON_VER_WARNING)
 endif
 
 check_conftest_imports:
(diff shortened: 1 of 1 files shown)


ajvb:
@hwine I'd like to understand why developers should be warned about this? As well, it feels unneeded currently as we do not have any external contributors, even outside of the team, as of now.

Also, I'd like to bump frost and zap to using Python 3.8 in production here soon, going to open a PR soon for a Ubuntu 20.04 version of our base deployment container.

ajvb:
Actually, thinking about it more, I'd rather we go for running tests against frost using python 3.6, 3.7, and 3.8.

hwine:
@ajvb -- it's being proposed because it cost me time. I developed on 3.8 and that broke when it tried to run my code on 3.6.

My guess is it will be hard to _always_ have all 3 environments in sync (dev, CI, prod), this gives a way to notify folks of changes.

I'll be happy to see all those environment updated -- but is that any reason to not have a safety net?

I'll agree that supporting non-Mozilla users could lead to conflicts, but as you noted, that case doesn't yet exist.

hwine:
> 
> 
> Actually, thinking about it more, I'd rather we go for running tests against frost using python 3.6, 3.7, and 3.8.

good, but different discussion -- opened #306 for it

ajvb:
@hwine 

>@ajvb -- it's being proposed because it cost me time. I developed on 3.8 and that broke when it tried to run my code on 3.6.

Ok totally.

>good, but different discussion -- opened #306 for it

Is it a different discussion? I don't like the idea of having to keep the `PRODUCTION_PYTHON` variable synced up with what is being ran in production without automation around updating it.

If CI tests python 3.6+, then python 3.8 specific code would fail in CI before making it to production.

hwine:
> If CI tests python 3.6+, then python 3.8 specific code would fail in CI before making it to production.

Agreed. But we didn't do that. We moved to 3.8 for CI without (imo) adequate notification that 3.6 is what matters for production.

To me, adding this warning (not error) is most visible when creating a venv, felt like the lowest friction way to address things. 

I _like_ the idea of getting on 3.8!!!  I wouldn't want to roll back to CI on 3.6 just to make things "consistent". The coordination of updating the three environments is already going to be challenging to find uninterrupted time for. Why create a "critical path" if we can avoid it?

> Is it a different discussion? I don't like the idea of having to keep the `PRODUCTION_PYTHON` variable synced up with what is being ran in production without automation around updating it.

I don't see us investing in automating upgrades across 3 environments. I see this action as a "checklist item" in the "run book" for doing the upgrade. Especially as more folks & teams get involved in using frost, the "explicit is better than implicit" guidance from `import this` seems worth it to me.

Makefile
```diff
@@ -7,6 +7,12 @@ AWS_PROFILE := default
 
 PYTEST_OPTS := ''
 
+PRODUCTION_PYTHON := 3.6
```
g-k, line 10: nit: seems weird to mention production python in the OS repo. Should we just check for versions we don't test against in CI?

hwine: agreed -- I'll switch to PYTHON_MIN_VERSION

g-k:
Thanks Hal!

Issue #306, opened 2020-07-17
Decide which python versions to support, and how to deal with that


hwine:
Actually, thinking about it more, I'd rather we go for running tests against frost using python 3.6, 3.7, and 3.8.

_Originally posted by @ajvb in https://github.com/mozilla/frost/pull/305#issuecomment-660346608_

hwine:
My concerns about this are how good our coverage is. I may not understand yet all the complexity yet, but it feels risky to do version bumps. Part of that is the classic issue of "testing the tests", and how brittle the ad-hoc systems frost integrates with are (e.g. metrics).

Trying to be more specific:
- our doctest coverage is only ~18%. To me, that suggests that a lot of our "testing" is running in production. We won't really know if we're good on a non-prod version until we try it there.
- related, for the 2 compatibility errors I had, only one was a compile time error. The 2nd was a runtime error in a `subprocess` call. That will be hard to mock.

If we do decide to pick up the extra burden, I would advocate for only supporting 2 versions: current prod and next prod. I don't see the value add of a "middle version". (Yes, it _should_ only cost extra test time, but I'd be more comfortable with an explicit value statement to justify the extra maintenance effort along the way.)

Finally, I'll suggest that we might want to make this a team wide decision. All the arguments for doing it for frost apply to our other active projects as well. It wouldn't help productivity to have each project have a disjoint set of supported python versions.

hwine:
Another thought around versions that just hit me is: Is it a reasonable assumption that frost _can_ be limited to running on a single version of python?

Noticing that:
- While we think of frost as a single framework, in practice we already invoke it on sub-trees of the repository. 
- Each system that is monitored has its own client setup, which could require specialized libraries. If a library hasn't been ported to a new version, do we want to hold _everyone_ back from benefits of upgrading? Or do we say "fork it - you're on your own"?

One approach to addressing this would be to redefine the system boundaries such that frost is the core, and perhaps does support 3 or more python versions. Each subtree would be responsible for ensuring it runs under an appropriate python, and is tested for that/those python versions. (I think that would be relatively easy with the way doctests integrates with pytest.)

(For reasons that aren't entirely germane to this discussion, I would love to have "sub trees" be usable outside the frost framework.  Or phrased differently, be able to easily wrap an existing set of tests into frost, perhaps just by importing that external package.)

ajvb:
Closed with #307

PR #307, opened 2020-07-22
Test on multiple versions of python

ajvb:
Run on Python 3.6+ in CI. Don't specify python version in pre-commit config.

Related to #304, #305, #306

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 python:
+  - "3.6"
+  - "3.7"
   - "3.8"
 sudo: required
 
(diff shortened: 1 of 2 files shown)


g-k:
I'm in favor of staying on 3.8 / latest once that's possible. We can package a docker image to make that easier and have people mount their custom tests.

.travis.yml
```diff
@@ -1,5 +1,7 @@
 language: python
 python:
+  - "3.6"
+  - "3.7"
```
hwine, line 4: I don't see the need for 3.7, since we don't intend to ever run that in production. If something worked on 3.6 & 3.8, we wouldn't want to fix 3.7. (A bit hypothetical, but basically a vote for 2 versions: current prod & next prod)

hwine:
> ...We can package a docker image to make that easier and have people mount their custom tests.

I think this is one of several options for a more explicit development story. I'm looking forward to chatting about those. :)

Issue #302, opened 2020-07-06
package a docker image


g-k:
It's weird that we:

* have both a repo and the pypi package: https://pypi.org/project/frost/ and 
* create a virtualenv for the install (e.g. in docker images)

We could run `frost` as an alias to `pytest` on a directories of tests or individual test files with an optional config file (and other pytest args passing through to pytest):

`frost --config config.yaml tests/`

It'd also let us bundle up tests and their related test resources as pytest plugins in the future: https://docs.pytest.org/en/stable/writing_plugins.html#writing-plugins

g-k:
f? @ajvb

ajvb:
@g-k I completely agree. I also think this would help with usability and user experience for those new to frost. Saying "Run the `frost` command" is a lot easier than saying "Run the `pytest` command with this selection of custom pytest cli args but ignore all these other ones" imo

g-k:
Renaming issue to package as a docker image. I was running into issues with outdated pytest and generally wrapping it without breaking existing APIs.

I think we can:

- [ ] package frost as a docker image
- [ ] build and publish it from CI (need creds from a mozilla docker admin)
- [ ] migrate existing users to the docker image
- [ ] then eventually swap the CLI if we want

The benefits would be:

* pin our Python and pytest versions
* default to not running all tests
* mount in user tests
* add extra non-python tools and not worry about platform specific changes

An additional consideration is getting this working nicely with VSCode since Hal and other users use it.

ajvb:
@g-k In regards to trouble with wrapping pytest, do you know if that would be resolved with upgrading pytest? I still really like the user experience story of having a `frost` command, even if we just keep it on the backlog for a while.

g-k:
I did run into the `ignore_tests` config requiring a newer version, but didn't want to try to update pytest with the CLI change.

I was able to run pytest programmatically and pass through args using the click packages, so I'll probably revisit that at some point. Ideally we make frost less magical, remove reporting and other plugins, weird conftest stuff, and just wrap the individual test running functionality sort of like:

```console
$ cat jsonlines-data.jsonl | front this-one-weird--included-test > output.jsonl # where output is a report or a list of actions to run in other tools (create bugs, file issues, etc.)
$ cat jsonlines-data.jsonl | front my-test.py > output.jsonl
$ # or run multiple tests
$ frost --config my-config-specifying-tests-to-run
```

g-k:
The docker image is busted.

[...]

PR #310, opened 2020-07-30
Fix 302 containerize

g-k:
refs: #302

Changes:
* Add a docker container based on debian buster slim
* update readme (NB: replaces old example output with a TODO) 
* adds CCI config to publish images as `mozilla/frost` (need to enable in CCI and add dockerhub creds)

r? @ajvb

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,29 @@
+# frost
+
+FROM python:3.8-slim-buster
+
+ENV PYTHONPATH $PYTHONPATH:/app
+ENV PYTHONUNBUFFERED 1
+
+RUN groupadd --gid 10001 app && \
+    useradd --uid 10001 --gid 10001 --shell /usr/sbin/nologin app
+RUN install -o app -g app -d /var/run/depobs /var/log/depobs
+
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+COPY meta_test_cache.py .
+COPY regressions.py .
+COPY service_report_generator.py .
+COPY severity.py .
+COPY bin bin
+COPY aws aws
+COPY gcp gcp
+COPY gsuite gsuite
+COPY heroku heroku
+COPY pagerduty pagerduty
```
ajvb, line 43: Why all of these `COPY` instructions rather than a `COPY * .`?

ajvb: Feel like this could easily get out of sync with new additions and could be kind of a pain to determine if everything we need is listed here.

g-k: yeah that + a `.dockerignore` is probably cleaner I'll update it

g-k:
Updated. Do we want to configure and test CCI before landing?

g-k:
ok I'll land this and we can get CI working later

Issue #316, opened 2020-08-03
CircleCI builds failing


hwine:
needs tuning to run not-on-local machine

PR #317, opened 2020-08-03
Make circleci work

hwine:
Tweak some syntax and don't try to run tests that will fail.

Fixes GH-316


Dockerfile
```diff
@@ -23,7 +23,7 @@ WORKDIR /app
 COPY requirements.txt .
 RUN pip install --upgrade --no-cache-dir -r requirements.txt
 
-COPY * .
+COPY * ./
```
g-k, line 26: ah TIL

PR #319, opened 2020-08-04
Replace use of 'rg' for CircleCI

hwine:
'rg' isn't part of the base images, and may not be on some dev's
machines. So switch to the old standby of 'grep' -- with a bunch more
options for the same behavior.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,13 +53,13 @@ jobs:
       - run:
           name: deploy to Dockerhub
           command: |
-            # publish docker image master and scan envs
+            # deploy docker image master based on scan envs
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
               docker login -u $DOCKER_USER -p $DOCKER_PASS
               docker tag "localhost/frost:latest" "${DOCKERHUB_REPO}:latest"
               docker push "${DOCKERHUB_REPO}:latest"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -88,6 +87,13 @@ workflows:
           filters:
             tags:
               # only upload the docker container on semver tags
-              only: /[0-9]\.[0-9]+\.[0-9]+/
+              only: /^v[0-9]\.[0-9]+\.[0-9]+/
             branches:
-              only: master
+              # filters appear to be OR'd together, so we can't specify just
+              # 'master' (as it tries to upload with every commit to
+              # master).
```
g-k, line 94: we usually deploy master as the `latest` tag

hwine: I'm not sure I follow this, so likely should be a separate bug. As configured now, an image is pushed to dockerhub whenever a tag matching a 3 level semver tag is applied (with a leading 'v'). It does not matter which branch the tag is applied to.

g-k:
Docker uses the tag "latest" as a default and projects usually publish
their default/master as the docker "latest" tag.

Previously published versions of frost don't have a leading v:

https://github.com/mozilla/frost/tags
https://pypi.org/project/frost/#history

we can change versioning schemes though

PR #321, opened 2020-08-05
Cleanup service report generator

ajvb:
* Removes a lot of unused code in `service_report_generator.py`
* Adds `resource_name` as a top-level parameter.

From looking at usage of this script and the resulting json files, this shouldn't be a breaking change. The markdown and csv outputs were used early in the project and haven't been for probably over a year.

As well, they were a pain to maintain for AWS and will only be worse for more services like GCP, GSuite, Heroku, etc.


ajvb:
@g-k :+1: sounds good!

I believe we are using:
```
pytest-json==0.4.0
pytest-metadata==1.8.0
```

are we using `pytest-cov==2.6.1` anywhere?

g-k:
I don't think we're using pytest-cov we should try removing it

ajvb:
Ah, it's used for the `coverage` make command.

PR #322, opened 2020-08-05
readme: add logo

g-k:
per @jvehent logo is licensed for public use


ajvb:
lol...

g-k:
Not sure why this isn't showing up in the preview, but maybe it'll work once landed?

hwine:
The URL doesn't refer to a branch, so it's looking on the default branch. The png won't be there until after it's merged.

_[aside: it'll be interesting to see if this "works right" with my next docs update -- I'm pulling the top level README.md into the docs. that should work since we're hosting on github.io]_

hwine:
oops, my bad - you _also_ can't use a relative URL there i.e. still not showing after merge.

PR #330, opened 2020-09-11
Bpitts/aws ec2 ebs piops

sciurus:
Cloudops does not want to use any provisioned iops volumes in EC2. This test will let us catch any places where we are.

In addition to the test itself, I've had to change the cache patch to make it more strict when converting strings to datetimes. The volume type `st1` was being incorrectly converted to `datetime.datetime(2020, 9, 1, 0, 0)`.

I'm unsyre if this will negatively affect tests for any other providers.

### DIFF
--- a/aws/ec2/helpers.py
+++ b/aws/ec2/helpers.py
@@ -292,6 +292,7 @@ def is_ebs_volume_encrypted(ebs):
     """
     return ebs["Encrypted"]
 
+
 def is_ebs_volume_piops(ebs):
     """
     Checks if the EBS volume type is provisioned iops
--- a/aws/ec2/helpers.py
+++ b/aws/ec2/helpers.py
@@ -292,6 +292,29 @@ def is_ebs_volume_encrypted(ebs):
     """
     return ebs["Encrypted"]
 
+def is_ebs_volume_piops(ebs):
+    """
+    Checks if the EBS volume type is provisioned iops
+
+    >>> is_ebs_volume_piops({'VolumeType': 'io1'})
+    True
+    >>> is_ebs_volume_piops({'VolumeType': 'standard'})
+    False
+    >>> is_ebs_volume_piops({})
+    Traceback (most recent call last):
+    ...
+    KeyError: 'VolumeType'
+    >>> is_ebs_volume_piops(0)
+    Traceback (most recent call last):
+    ...
+    TypeError: 'int' object is not subscriptable
+    >>> is_ebs_volume_piops(None)
+    Traceback (most recent call last):
+    ...
+    TypeError: 'NoneType' object is not subscriptable
+    """
+    return ebs["VolumeType"].startswith("io")
+
 
 def is_ebs_snapshot_public(ebs_snapshot):
     """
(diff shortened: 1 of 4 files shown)


ajvb:
This looks great. Looks like the linter (https://github.com/psf/black) is failing in CI.

g-k:
> I'm unsyre if this will negatively affect tests for any other providers.

Should we throw that behind a feature flag e.g. env var `STRICT_ISODATES=1` for a gradual migration or just land it and update whatever fails?

ajvb:
@g-k I'm not sure. For our automated runs of Frost, this should have no impact, as we never use the frost resources cache. This could mess with folks local cache, but I'm unsure on the impact outside of potentially messing with tests that are related to time, i.e. https://github.com/mozilla/frost/blob/master/aws/iam/test_iam_access_key_is_old.py

Are there other cases that I may be missing?

g-k:
> For our automated runs of Frost, this should have no impact, as we never use the frost resources cache.

Oh good point.

> Are there other cases that I may be missing?

I don't think so.


Yeah, let's just land this. I'm fine asking people to clear their local cache.

aws/ec2/helpers.py
```diff
@@ -293,6 +293,30 @@ def is_ebs_volume_encrypted(ebs):
     return ebs["Encrypted"]
 
 
+def is_ebs_volume_piops(ebs):
+    """
+    Checks if the EBS volume type is provisioned iops
+
+    >>> is_ebs_volume_piops({'VolumeType': 'io1'})
```
g-k, line 300: oh nice! +1 for comprehensive doctests

aws/ec2/helpers.py
```diff
@@ -293,6 +293,30 @@ def is_ebs_volume_encrypted(ebs):
...
+    ...
+    KeyError: 'VolumeType'
+    >>> is_ebs_volume_piops(0)
+    Traceback (most recent call last):
+    ...
+    TypeError: 'int' object is not subscriptable
+    >>> is_ebs_volume_piops(None)
+    Traceback (most recent call last):
+    ...
+    TypeError: 'NoneType' object is not subscriptable
```
g-k, line 315: Note to self: if we start typehinting e.g. with typeddict we can reduce the number of these types to unit test

aws/iam/meta_test_resources.py
```diff
@@ -7,17 +7,17 @@ def test_iam_users():
...
             "UserId": "H0BBIHMA0CZ0R0K0MN00C",
             "UserName": "tigerone",
             "__pytest_meta": {"profile": "example-account", "region": "us-east-1"},
         },
         {
             "Arn": "arn:aws:iam::123456789012:user/calvin",
-            "CreateDate": parse("1985-11-18T00:01:10+00:00", ignoretz=True),
-            "PasswordLastUsed": parse("2008-01-09T20:43:00+00:00", ignoretz=True),
+            "CreateDate": isoparse("1985-11-18T00:01:10+00:00"),
+            "PasswordLastUsed": isoparse("2008-01-09T20:43:00+00:00"),
```
g-k, line 20: :tada: I bet there are some old issues I can close out related to this.

g-k (approved):
r+

ajvb (approved):

PR #332, opened 2020-09-14
elb: add test for http desync setting

g-k:
Assuming https://docs.google.com/document/d/1yKLDxZ9S9N456ZPh4qEM0GY2DrMZqsqG69ByCSELjzE/edit is approved this PR adds a test to check whether ELBs are deployed in HTTP Desync strictest mode. I have a doctest for the test itself, but have not functional tested that the resources change fetches the right data. 

We would need to add exemptions for current ELBs, but shouldn't need to for new ELBs.

Alternatively, we could change this to a negative test to make sure we're not running in `monitor` mode anywhere and not add exemptions for existing services.

r? @ajvb 

cc @sciurus @jbuck

### DIFF
--- a/aws/elb/resources.py
+++ b/aws/elb/resources.py
@@ -1,7 +1,7 @@
 from conftest import botocore_client
 
 
-def elbs():
+def elbs(with_tags=True):
     """
     http://botocore.readthedocs.io/en/latest/reference/services/elb.html#ElasticLoadBalancing.Client.describe_load_balancers
     """
@@ -11,6 +11,9 @@ def elbs():
         .flatten()
         .values()
     )
+
+    if not with_tags:
+        return elbs
 
     elbs_with_tags = []
     for elb in elbs:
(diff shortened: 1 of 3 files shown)


ajvb (approved):
>but have not functional tested that the resources change fetches the right data.

We should test this prior to merging. Is enabling this simple or are there extra steps behind it? I usually just create test resources in the dev account.

jbuck:
> We would need to add exemptions for current ELBs, but shouldn't need to for new ELBs.
> 
> Alternatively, we could change this to a negative test to make sure we're not running in `monitor` mode anywhere and not add exemptions for existing services.

Maybe test if ELB created at > 2020-09-01 it should be strict, else ELB should be strict or defensive? I guess it depends how aggressively we want to turn it on everywhere

g-k:
Updated to add the time cutoff and functional tested on cloud svcs dev.

PR #335, opened 2020-09-15
Add test for checking specific versions of GKE

ajvb:
For assistance with new vulns like https://cloud.google.com/kubernetes-engine/docs/security-bulletins#gcp-2020-012

### DIFF
--- a/gcp/compute/test_only_allowed_gke_versions.py
+++ b/gcp/compute/test_only_allowed_gke_versions.py
@@ -1,0 +1,32 @@
+import pytest
+
+from gcp.compute.resources import clusters
+from conftest import gcp_client
+
+
+@pytest.fixture
+def allowed_gke_versions(pytestconfig):
+    return pytestconfig.custom_config.gcp.allowed_gke_versions
+
+
+@pytest.mark.gcp_compute
+@pytest.mark.parametrize(
+    "cluster", clusters(), ids=lambda c: c["name"] if isinstance(c, dict) else None
+)
+def test_only_allowed_gke_versions(cluster, allowed_gke_versions):
+    """
+    Tests if GKE version is within allowed list of GKE versions.
+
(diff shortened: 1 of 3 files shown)


gcp/compute/test_only_allowed_gke_versions.py
```diff
@@ -0,0 +1,32 @@
...
+)
+def test_only_allowed_gke_versions(cluster, allowed_gke_versions):
+    """
+    Tests if GKE version is within allowed list of GKE versions.
+
+    Useful for checking upgrade status after a vulnerability is released, as in:
+        - https://cloud.google.com/kubernetes-engine/docs/security-bulletins#gcp-2020-012
+    """
+    assert (
+        cluster["currentMasterVersion"] in allowed_gke_versions
```
sciurus, line 24: Eventually we'll want to run this check on an ongoing automated basis. Won't it then start raising false positives once Google releases versions newer than our allowed versions and we are auto-upgraded? It's probably better to parse the version string and have a minimum version for each minor version. I.E fail for 1.15.12-gke.19 but pass for both 1.15.12-gke.20 and 1.15.12-gke.21.

ajvb: For the automated runs, it might be better to use https://github.com/mozilla/frost/blob/master/gcp/compute/test_gke_version_up_to_date.py which uses https://cloud.google.com/sdk/gcloud/reference/container/get-server-config to check for valid versions. GCP keeps that list up-to-date so we can get notified of old versions. I'm thinking that this test will be more useful for manually checking when there is a specific sec vuln released.

sciurus: Gotcha. That seems reasonable.

g-k (approved):
r+ lgtm

Issue #311, opened 2020-07-30
default to not collecting test data


g-k:
Currently we have kind of insane `--ignore` strings, which is confusing and defaults to frost trying to slowly collect resources on each run. Proposal:

- [x] upgrade to pytest 4.3.0+ https://docs.pytest.org/en/stable/changelog.html#pytest-4-3-0-2019-02-16 (and dependent plugins)
- [x] add `collect_ignore_glob` to conftest.py https://docs.pytest.org/en/stable/reference.html#globalvar-collect_ignore_glob
- [x] update docs to remove `--ignore` strings and require explicit test inclusion

PR #336, opened 2020-09-15
default to not collecting any tests

g-k:
fix: #311

I believe all our current frost jobs explicitly include each test to run.

r? @ajvb 

edit: CI is failing; fixed


ajvb (approved):
:tada:

Issue #113, opened 2018-05-09
Allow pytest-services to be pip installed via PyPI


JohnVonNeumann:
I would assume there to be obvious benefits in this, if entirely possible I wouldn't mind being able to just install this the way I'd install my other packages, without really caring about your dependencies.

hwine:
I'm not sure how this would work. The entire repo is required - there aren't any executables or libraries to include elsewhere. It is not a Python package or module at present.

Can you describe the usage scenario you envision? I.e. what are you doing after the ``pip install``?

g-k:
Hey @JohnVonNeumann, thanks for checking this out!

For now, the easiest way to run these would be to clone the repo, add a config, and run the tests you're interested in.

We originally planned to refactor this into pytest plugins (#3 and #4) and still might do that, but as Hal pointed out we decided to open source it as a bunch pytests for now.

The plugin based workflow would might look like:

1. have a pytest unit tests for you python application
1. add relevant pytest-services plugins as pip-installeable packages for your cloud provider, OS, monitoring services, etc. 
1. test them in CI or AWS config

Is that the kind of scenario you had in mind?

PR #341, opened 2020-09-16
Fix 302 add cli bin

g-k:
refs #113 #302

Changes:

* Add a pypi package with CLI bin cmd `frost` ... it doesn't do anything
  * subcommand `frost test` aliased to `pytest -s --debug-calls` run from the frost test root 
  * subcommand `frost list` to print out test file paths (since we default to ignoring all of them)
* Update the readme to use the `frost` command instead of docker (we never published the docker image and would have to mount tests in)
* fix the `awsci` make target to run aws tests

NB: **running tests via `pytest` is unchanged**

Caveats:

* heroku tests are blocked on being able to install the package: https://github.com/hwine/python-herokuadmintools/issues/1
* the test files still live outside the frost package directory

This should:

* provide a single entrypoint independent of pytest for future iteration
* make it easier to split out the fetching and caching functionality as separate subcommands or e.g. `make setup_gsuite` as a distinct subcommand
* provide a way to package and distribute tests without cloning this repo

r? @ajvb

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 import setuptools
 
 with open("README.md", "r") as fh:
--- a/setup.py
+++ b/setup.py
@@ -1,19 +1,20 @@
 #!/usr/bin/env python
 
 import setuptools
+from frost import SOURCE_URL, VERSION
 
 with open("README.md", "r") as fh:
     long_description = fh.read()
 
 setuptools.setup(
     name="frost",
-    version="0.3.1",
+    version=VERSION,
     author="Firefox Operations Security Team (foxsec)",
     author_email="foxsec+frost@mozilla.com",
     description="tests for checking that third party services the Firefox Operations Security or foxsec team uses are configured correctly",
     long_description=long_description,
     long_description_content_type="text/markdown",
-    url="https://github.com/mozilla/frost",
+    url=SOURCE_URL,
     license="MPL2",
     packages=setuptools.find_packages(),
     classifiers=[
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -1,25 +1,33 @@
...
     classifiers=[
         "Natural Language :: English",
         "License :: OSI Approved :: Mozilla Public License 2.0 (MPL 2.0)",
         "Operating System :: OS Independent",
-        "Programming Language :: Python :: 3.6",
-        "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
     ],
-    python_requires=">=3.6",
+    python_requires=">=3.8",
```
ajvb, line 31: This will make https://github.com/mozilla/frost/blob/master/Makefile#L10-L14 kind of strange. Can we remove the min python version checking in Makefile now that this is here?

ajvb (approved):
:tada: this is awesome! Thanks @g-k !

I'd like to remove the min python version checking in `Makefile` now that this is in place.

As well, I opened https://github.com/mozilla/frost/issues/344 for pushing of new tagged versions to PyPI via Travis.

setup.py
g-k: I think so, do we have anything using Python 3.6 still?

ajvb: Nope, Jenkins runs are all on 3.8 now.

g-k: updated

g-k:
Nice!

g-k:
fixes: #113 #302

g-k:
Still need to tag and deploy.

Issue #117, opened 2018-05-21
Agree on persistent caching pattern


hwine:
Some systems, notably GitHub, take significant time (hours) to acquire all the base data. For that reason,  it would be good to have a way to persist the cache between runs.

To Do:
- [x] Propose strategy
- [ ] Get agreement
- [ ] Implement (in GitHub)

g-k:
@hwine sounds good! +1 on per service caches (didn't realize you could do per directory cachedirs but that makes sense) and mounting the directory. This would also fix #2 (e.g. by running `rm .cache/v/aws/iam` first).

In general I'm not super happy about having pytest fetch data directly. The main motivation for doing that was to not fetch any more data than the test required. We should be able to decouple the API clients and pytest.

Some other things that might be good to update with our caching:

* content addressable hashes (e.g. sha256 of data to cache). The downside is it would be less clear what's in the cache, but it would fix the cache key escaping nonsense (and file name too long errors) and https://github.com/mozilla-services/pytest-services/issues/105 too assuming the AWS profile or ARN shows up somewhere in the data to cache. 
* fs file limit workarounds e.g. using the first two chars of hash as a subdirectory git-style
* fix the bug where deserialized datetimes read from cache were not `datetime.datetime`'s like they are in when fetched uncached

ajvb:
+1 to all of this.

I think it would also be nice to add the ability to **not cache** during a pytest run. The production runs for AWS always bust the cache and it would be made faster if they didn't write a bunch of stuff to disk unnecessarily

>In general I'm not super happy about having pytest fetch data directly.

I don't disagree with this, but what is the alternative?

g-k:
> to add the ability to not cache during a pytest run

I thought we had a flag for that, but that'd be easy to add.

In general, it'd be good to decouple reading and writing from cache too e.g. so we could mount github data from one directory as read only then write it to a different directory and only update the cache if the fetching succeeds.

> the alternative?

Have separate scripts that receive or prefetch data to populate the cache first or define a protocol for tests to say "I want this data" and have those scripts or services fetch it or both.

ajvb:
>I thought we had a flag for that, but that'd be easy to add.
>
>In general, it'd be good to decouple reading and writing from cache too e.g. so we could mount github data from one directory as read only then write it to a different directory and only update the cache if the fetching succeeds.

+1

[...]

Issue #11, opened 2018-01-24
Remove wrapped values and data processing from client.py


g-k:
refs: https://github.com/mozilla-services/pytest-services/pull/9

The client wrapping results creates extra lists in `extract_key` and other methods and is a weird non-pythonic API to learn.

A library of helper functions would be clearer and more explicit.

g-k:
@ajvb what do you think?

ajvb:
@g-k This is interesting. I do agree with you that it's not very pythonic, but at the same time I like the API you created within `BotocoreClient`. It could probably be made more "friendly" though.

g-k:
@ajvb If you get a chance to write something up, I'd be interested in how to make it more friendly. I think there are libraries for managing wrapped things or we stick things in a pandas dataframe.

g-k:
Something that behaves like lenses or zippers from FP might be good fit though wouldn't be pythonic.

ajvb:
👍 - I'm not really sure how to make it more friendly, which really means I'm not sure how to restructure the abstraction layer to get rid of the wrapped values. The current `extract_key`, `flatten`, `values` work flow is really nice and flexible once I understood it. Don't have much experience using dataframes in pandas so no opinion there.

ajvb:
The current model has been working great now for almost 3 years. I'm going to close this for now, can re-open if we want to address this again in the future.

PR #342, opened 2020-09-16
Prototype update for SNS tests in #334

smarnach:
This prototype encapsulates extracting region and profile in a new method on the `BotocoreClient` to allow the client code to look reasonable. It's not quite what I had in mind, but with the current structure it's the best I could come up with in an hour.

I also factored out the results list into its own class. It was a bit weird that the client stored the results in itself and then modified itself with `extract_key()` and `flatten()`. I simply moved that to a separate class, so subsequent calls to the client don't overwrite the previous results.

The test is now working and passing, but it's terribly slow. I tried it for the dev account, and it makes some 3,000 API calls, so you get rate limited and the client waits and retries. When I limit it to 300 subscriptions, it returns quite quickly, but running it for all subscriptions takes ages.

If the general approach seems reasonable, at least as an incremental improvement over the current situation, I can factor out the changes in the client in a separate PR and clean them up a bit.

Related: #334

### DIFF
--- a/aws/client.py
+++ b/aws/client.py
@@ -103,6 +103,64 @@ def cache_key(call):
     )
 
 
+def get_single_region(
+    service_name,
+    method_name,
+    call_args,
+    call_kwargs,
+    cache,
+    profile,
+    region,
+    result_from_error=None,
+    debug_calls=False,
+    debug_cache=False,
+):
+    """
+    Fetches AWS API JSON responses for a single profile and region.
+    """
+    call = default_call._replace(
+        profile=profile,
+        region=region,
+        service=service_name,
+        method=method_name,
+        args=call_args,
+        kwargs=call_kwargs,
+    )
+
+    if debug_calls:
+        print("calling", call)
+
+    result = None
+    if cache is not None:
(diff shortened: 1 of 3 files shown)


smarnach:
I'll update this PR once #334 is merged – probably next week.

g-k:
:+1: 

> I also factored out the results list into its own class. It was a bit weird that the client stored the results in itself and then modified itself with extract_key() and flatten(). I simply moved that to a separate class, so subsequent calls to the client don't overwrite the previous results.

That'd be great! I used to write clojure and wanted something like https://clojuredocs.org/clojure.core/-%3E or a fluent interface, but it turned out badly (strange and incomplete). refs: https://github.com/mozilla/frost/issues/11

I pulled out a separate frost CLI entrypoint recently in #341 and adding a new subcommand e.g. `frost fetch` might provide the flexibility to try a saner client (e.g. `boto3`) and caching (refs #117) without having to deal with pytest.

smarnach:
I'll clean up the client changes and file them as a new PR against master.

smarnach:
Draft PR opened: #356

PR #343, opened 2020-09-17
Aws r53 tests

erkolson:
Initial pass at route53 tests


aws/route53/test_route53_zones.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+
+from aws.route53.resources import zones, cnames
+
+
+@pytest.mark.parametrize(
+    "cnames", cnames(),
```
smarnach, line 7: You should pass some `ids` parameter here, since otherwise failing tests are only identified by something like "cnames-123", which isn't really helpful.

aws/route53/resources.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def cnames():
+    records = []
+    zone_list = zones()
+
+    for zone in zone_list:
+        zone_id = zone["Id"].split("/")[2]
+        zone_records = (
+            botocore_client.get(
```
smarnach, line 23: This will iterate through all regions and profiles, which can cause problems. Since Route 53 is a global service, regions aren't a problem here, and we apparently don't use multiple profiles either. The fix is to pass at least the `profiles` parameter as Bob did in #334.

aws/route53/resources.py
```diff
@@ -0,0 +1,32 @@
...
+        botocore_client.get("route53", "list_hosted_zones", [], {})
+        .extract_key("HostedZones")
+        .flatten()
+        .values()
+    )
+
+
+def cnames():
+    records = []
+    zone_list = zones()
```
smarnach, line 18: I suggest inlining this variable.

aws/route53/resources.py
ajvb: re: regions iteration - This won't iterate through regions since https://github.com/mozilla/frost/pull/343/files#diff-5fc95b7e7fea5fe0334c6dffe5abedb8R14 was added in this PR.

ajvb (commented):
Looks like you are missing a `__init__.py` in `aws/route53/` (via the CI error)

aws/route53/resources.py
erkolson: Is the profile bit needed since the route53 api calls do not iterate through regions?

aws/route53/test_route53_zones.py
```diff
@@ -0,0 +1,11 @@
...
+
+from aws.route53.resources import zones, cnames
+from helpers import get_param_id
+
+
+@pytest.mark.parametrize(
+    "cnames", cnames(), ids=lambda record: get_param_id(record, "Name")
+)
+def test_route53_cnames(cnames):
+    assert int(cnames["TTL"]) >= 600, f"TTL is {cnames['TTL']}"
```
ajvb, line 11: I'd like to see a docstring for this test explaining why we are enforcing `600` here. As well, in the assert message, it might be nice to have something like `TTL is expected to be 600 or greater, but it is X` or something similar.

aws/route53/resources.py
ajvb: I think it's ok to not include it. @smarnach is correct in that there would be a problem if multiple profiles are used, but this is true with a number of our tests. Current plan is to deprecate allowing multiple AWS profiles in a single run.

ajvb (approved):

sciurus (approved):

Issue #323, opened 2020-08-05
Sort out planned docker tagging


hwine:
_Originally posted by @g-k in https://github.com/mozilla/frost/pull/319#issuecomment-669439863_

> Docker uses the tag "latest" as a default and projects usually publish their default/master as the docker "latest" tag.

> Previously published versions of frost don't have a leading v:

> https://github.com/mozilla/frost/tags
> https://pypi.org/project/frost/#history

> we can change versioning schemes though

It's not clear if it's really doing that. @gguthe has used the process elsewhere with success, so hoping he can sort out.

PR #345, opened 2020-09-17
remove circleci

g-k:
fixes: #323 (in that we aren't doing it)

We're only going to use PyPI to distribute frost, so the CCI docker build is redundant and never published to dockerhub.

We'll need to remove this repo from CCI.

r? @ajvb


hwine:
fwiw, I "removed" circle ci from the [required checks](https://github.com/mozilla/frost/settings/branch_protection_rules/1766465), so it won't block merging

ajvb (approved):

PR #347, opened 2020-09-18
Remove regression marking

### LINKED ISSUE #346: Consider removing regressions
https://github.com/mozilla/frost/blob/master/regressions.py and the corresponding config section.

I don't think we've used this. @g-k What do you think about us removing it?

ajvb:
Fixes #346

Removes the concept of marking a failing test result as a "regression". In practice, this requires a lot of manual effort with not enough value. It is much easier to look at test results as a time series or to search for bug's raised by Frost.


g-k (approved):
r+

PR #358, opened 2020-09-25
Check an S3 bucket should have a life cycle policy

bqbn:


aws/s3/test_s3_bucket_has_life_cycle_policy.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+from aws.s3.resources import s3_buckets, s3_bucket_lifecycle_configuration
+
+
+@pytest.mark.s3
+@pytest.mark.parametrize(
+    ["s3_bucket", "lifecycle_configuration"],
+    zip(s3_buckets(), s3_bucket_lifecycle_configuration()),
+)
```
ajvb, line 10: Can we add an `ids` param to the `parametrize` call? For ones like this where we zip() two resource calls together, we've done something like this:

test:
https://github.com/mozilla/frost/blob/047e3381581da2b940d0718c691db53e689e18e1/aws/iam/test_iam_user_without_mfa.py#L8-L12

helper:
https://github.com/mozilla/frost/blob/047e3381581da2b940d0718c691db53e689e18e1/aws/iam/helpers.py#L151-L158

bqbn: @ajvb This is fixed. Please review it again.

ajvb (approved):

PR #359, opened 2020-09-25
Check RDS storage type is not io1

bqbn:


aws/rds/test_rds_db_instance_storage_type_not_io1.py
```diff
@@ -0,0 +1,14 @@
...
+
+from aws.rds.helpers import get_db_instance_id
+from aws.rds.resources import rds_db_instances_with_tags
+
+
+@pytest.mark.rds
+@pytest.mark.parametrize(
+    "rds_db_instance", rds_db_instances_with_tags(), ids=get_db_instance_id,
+)
+def test_rds_db_instance_storage_type_not_io1(rds_db_instance):
```
ajvb, line 11: @bqbn Could we add a docstring here with a short description as to why we would want to check for this?

ajvb (approved):
LGTM! Would just like a short description for the test.

aws/rds/test_rds_db_instance_storage_type_not_io1.py
```diff
@@ -0,0 +1,14 @@
...
+    assert (
+        rds_db_instance["StorageType"] != "io1"
```
sciurus, line 13: @bqbn this will only catch the first-generation of provisioned iops volumes. AWS introduced io2 recently, which we'd want to flag as well. In #330 I checked for `.startswith("io")` to try to handle all current and future cases.

sciurus (changes requested):

aws/rds/test_rds_db_instance_storage_type_not_io1.py
bqbn: Done.

aws/rds/test_rds_db_instance_storage_type_not_io1.py
bqbn: Done.

sciurus (approved):

Issue #329, opened 2020-09-11
Handle disabled APIs for GCP


ajvb:
If APIs are disabled in GCP, tests will currently error out like so:

```
==================================== ERRORS ====================================
___ ERROR collecting frost/gcp/sql/test_sql_instance_private_ip_required.py ____
gcp/sql/test_sql_instance_private_ip_required.py:8: in <module>
    "sql_instance", instances(), ids=lambda instance: instance["name"]
gcp/sql/resources.py:5: in instances
    return gcp_client.list("sqladmin", "instances", version="v1beta4")
gcp/client.py:90: in list
    self._list(product, subproduct, version, results_key, call_kwargs)
gcp/client.py:112: in _list
    results = self._get_list_results(
gcp/client.py:149: in _get_list_results
    results = self._list_all_items(api_entity, call_kwargs, results_key)
gcp/client.py:159: in _list_all_items
    resp = request.execute()
venv/lib/python3.8/site-packages/oauth2client/_helpers.py:133: in positional_wrapper
    return wrapped(*args, **kwargs)
venv/lib/python3.8/site-packages/googleapiclient/http.py:838: in execute
    raise HttpError(resp, content, uri=self.uri)
E   googleapiclient.errors.HttpError: <HttpError 403 when requesting https://sqladmin.googleapis.com/sql/v1beta4/projects/<projectname>/instances?alt=json returned "Cloud SQL Admin API has not been used in project XXXX before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/sqladmin.googleapis.com/overview?project=XXXX then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.">
```

I'm not sure if this is desired. It might be preferable to print a descriptive message and instead skip the test. AIUI if an API is disabled like so, then it can be assumed that there are no resources to test.

ajvb:
Fixed with #361

PR #361, opened 2020-09-29
Rework GCP code to support folder id arg

### LINKED ISSUE #339: Improve GCP UX
* Should work similar to the gcloud libraries / CLI in regards to project / org selection
* Support scanning all projects in an org and all projects in a folder (code for the later in foxsec-automation)

ajvb:
Reworks the GCP client, resources, and test code to add support for
passing in a Folder ID and running tests against all of the projects
within that folder.

Alongside that, this includes some stability fixes to the client code,
creates a new "frost" named cli arg group, and removes caching from the
GCP client.

Fixes #339 and #329

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,5 +1,4 @@
 import argparse
-from sys import exit
 import datetime
 
 import pytest
@@ -101,9 +100,9 @@ def pytest_configure(config):
     project_id = config.getoption("--gcp-project-id")
     folder_id = config.getoption("--gcp-folder-id")
     if project_id is not None and folder_id is not None:
-        # TODO: Probably a better way to cleanly exit than `exit(1)`?
-        print("--gcp-project-id and --gcp-folder-id are mutually exclusive arguments")
-        exit(1)
+        raise Exception(
+            "--gcp-project-id and --gcp-folder-id are mutually exclusive arguments"
+        )
 
     organization = config.getoption("--organization")
 
--- a/conftest.py
+++ b/conftest.py
@@ -1,4 +1,5 @@
 import argparse
+from sys import exit
 import datetime
 
 import pytest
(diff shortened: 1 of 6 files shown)


conftest.py
```diff
@@ -93,6 +99,12 @@ def pytest_configure(config):
     )
 
     project_id = config.getoption("--gcp-project-id")
+    folder_id = config.getoption("--gcp-folder-id")
+    if project_id is not None and folder_id is not None:
+        # TODO: Probably a better way to cleanly exit than `exit(1)`?
```
g-k, line 104: Maybe `raise Exception("Invalid arguments. Cannot use --gcp-fol.. and --gcp-proj..")`

conftest.py
```diff
@@ -23,51 +24,56 @@
 
 
 def pytest_addoption(parser):
-    parser.addoption(
+    frost_parser = parser.getgroup("Frost", "Frost's custom arguments")
```
g-k, line 27: :+1: this will help decoupling things if we want to move more logic to the top level click CLI parser

conftest.py
```diff
@@ -106,9 +118,8 @@ def pytest_configure(config):
 
     gcp_client = GCPClient(
         project_id=project_id,
-        cache=cache,
```
g-k, line 109: :tada:

gcp/client.py
```diff
@@ -1,81 +1,160 @@
...
+
 from apiclient.discovery import build as build_service
+from apiclient.errors import HttpError
+
+warnings.filterwarnings(
+    "ignore", "Your application has authenticated using end user credentials"
+)
+
 
+def get_all_projects_in_folder(folder_id=None):
```
g-k, line 11: nit: a doctest would be handy to show what's going on here

gcp/client.py
```diff
@@ -98,47 +186,28 @@ def _list(
...
+
         service = self._service(product, version)
 
         api_entity = getattr(service, subproduct.split(".")[0])()
         for entity in subproduct.split(".")[1:]:
             api_entity = getattr(api_entity, entity)()
 
         if self.debug_calls:
-            print("calling", api_entity)
+            print(
```
g-k, line 204: TODO: use the logging module and remove the random debug flags

as a new issue

gcp/compute/resources.py
```diff
@@ -14,34 +14,43 @@ def instances():
...
+            "projects.locations.clusters",
+            results_key="clusters",
+            call_kwargs={"parent": "projects/{}/locations/-".format(project_id)},
+        )
+    return results
 
 
 def networks_with_instances():
+    allInstances = instances()
+    inUseNetworks = []
```
g-k, line 30: nit: snake_case would be more pythonic but if the API var names are camelCase that might be clearer

g-k (approved):
r+

gcp/client.py
g-k: filed as https://github.com/mozilla/frost/issues/362

gcp/client.py
ajvb: I'm happy to do so, but I'm not sure a doctest would work since this is making a number of calls to GCP. I'm happy to write a unit test for this with mocks if desired?

g-k:
Ugh, no that sounds gross.

ajvb:
lol word.

PR #368, opened 2020-10-01
Upgrade google_api_python_client

ajvb:


g-k (approved):
r+

g-k:
Should we re-enable dependabot with a weekly or monthly cadence?

ajvb:
@g-k Yeah! I think that would be good now that frost is getting more attention again.

PR #369, opened 2020-10-01
Doc update: Add a getting started section.

Micheletto:
The Sphinx documentation build is part of this PR, which makes it seem like much more is going on.


docs/GettingStarted.rst
```diff
@@ -0,0 +1,86 @@
...
+================
+
+Clone the repo::
+
+   git clone git@github.com:mozilla/frost.git
+
+Create a virtual environment::
+
+   cd frost
+   python3 -m venv venv
```
hwine, line 22: The preferred way for all these commands is to use the `Makefile`

If you do `make` in a fresh checkout, you get instructions that will lead you to running the tests.

That's a "convention" for secops repos, but not documented (or adhered to, completely)

docs/GettingStarted.rst
```diff
@@ -0,0 +1,86 @@
...
+   . venv/bin/activate
+
+Upgrade pip::
+
+   pip install --upgrade pip
+
+Install Requirements::
+
+   pip install -r requirements.txt
+
```
hwine, line 35: One thing not in the `Makefile` is installing 'frost' in development mode, so you can use the `frost` command as well. To do that,  run:
```python
pip install -e .
```

docs/GettingStarted.rst
```diff
@@ -0,0 +1,86 @@
+.. raw:: html
+
+   <!-- This Source Code Form is subject to the terms of the Mozilla Public
+      - License, v. 2.0. If a copy of the MPL was not distributed with this
+      - file, You can obtain one at https://mozilla.org/MPL/2.0/. -->
+
+===============
+Getting Started
+===============
+Get started by installing Frost and running some pre-built tests.
```
hwine, line 10: What flavor of "getting started" are you intending? Getting started using or developing or ???

The reason I ask is that "using" is covered fairly well in the repo's README file. There's a bunch of new stuff there about using the `frost` command. aiui, that's the path forward for usage. (I'll try to add the readme to these docs -- it's a trick as the readme is markdown, and the docs are rsT.)

I'll assume the intent of this doc is "getting started developing". So adding that ("developing") to the title will also make that distinction clear in the table of contents.

docs/GettingStarted.rst
```diff
@@ -0,0 +1,86 @@
...
+Frost uses the `boto 3 <https://github.com/boto/boto3/>`_ library to communicate with AWS. Prior to running tests you must `configure <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/quickstart.html#configuration>`_ your environment authentication credentials to work with boto3, and then authenticate to AWS.  See the AWS `documentation <https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html>`_ for assistance.
+
+Pick a test
+-----------
+
+The tests for AWS live under the associated services directories in the aws directory.
+
+Example run::
+
+   pytest aws/elb/test_elb_instances_attached.py
```
hwine, line 51: Yes, AND there are often options you want to specify.  Maybe not appropriate in the getting started beyond a note to "see X for available options".

I'm not up on the details of running AWS tests, but I believe there are ways to specify which account(s) to examine.

hwine (changes requested):
lgtm -- the main change is to use the `make` commands to create the venv & install the requirements.

The rest of the comments are suggestions.

docs/GettingStarted.rst
Micheletto: Before writing this I re-read the Sphinx docs, and the foxsec-automation docs, and did not come back to the README here.  I'll check back on this tomorrow, but I'll probably just scrap this whole PR.

sciurus: Does this suggest there are too many places things are documented?

@hwine do you think we need both an extensive readme and the sphinx docs, or could we merge most of the readme into sphinx and link to it there?

hwine: We'll chat at today's session, but tl;dr I plan to include the README in the rendered docs. (some folks cringe at 2-screenful functions, but think 700 line docs don't need refactoring) :D :D

Micheletto:
Closing this.

PR #373, opened 2020-10-02
pin python dependencies

### LINKED ISSUE #367: Consider pinning python versions (including tooling like pip)
We just experienced an unexpected production breakage when a CI build may have used the new `pip` resolver. (i.e. broken by upstream software upgrade)

Currently, none of the build tooling is pinned. As we spread to more users, the blast radius of sudden failures expands. 

One way to address only bump versions as part of a new release (that's been tested)

g-k:
Not best practice, but maybe it'll fix #367

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ with open("README.md", "r") as fh:
     long_description = fh.read()
 
 install_requires = [
-    line.split("==")[0] for line in open("requirements.txt", "r") if ".git" not in line
+    line for line in open("requirements.txt", "r") if ".git" not in line
 ]
 
 setuptools.setup(


setup.py
```diff
@@ -7,7 +7,7 @@
     long_description = fh.read()
 
 install_requires = [
-    line.split("==")[0] for line in open("requirements.txt", "r") if ".git" not in line
+    line for line in open("requirements.txt", "r") if ".git" not in line
```
hwine, line 10: I really don't like this approach -- it's magic, and makes the production environment differ from dev, potentially leading to hard-to-debug conditions.

Since we've eliminated the repo reference from requirements.txt, I don't think this PR is needed at all.

Two thoughts if you do want to proceed:
1. remove requirements from setup.py entirely, in favor of specifying as much as possible in `pyproject.toml` as proposed in #372 
2. annotating this with a huge, ugly comment, so it's very obvious that magic is occuring.

hwine (changes requested):
My $0.02

PR #377, opened 2020-10-05
Publish documentation from Travis to gh-pages branch.

smarnach:
This enables publishing the documentation to a separate branch, so the changes in the generated files don't clutter up pull request diffs, and we don't need to manually add the changes anymore. I didn't need to add the `githubpages` extension for Sphinx, since [it is already enabled](https://github.com/hwine/sphinx-markdown/blob/bae170f8adf1d2b669f1f6e845862f6fa58f3a5c/docs/conf.py#L34). Furthermore, `docs/_build` is already in `.gitignore`, so the generated files will be ignored once we removed them from the repo.

There are a few further steps required to make this work:

* Create a Github deploy key with write permissions and add it to Travis as `GITHUB_TOKEN`.
* Change the GitHub Pages source to the root directory of the `gh-pages` branch.

These steps need to be done by someone with the power to do so (i.e. not me).

Once we are publishing from the new source, I will remove the documentation in the master branch from the repo.

This PR also fixed the `.PHONY` target in the Makefile. The old target didn't have any declared dependencies, but instead a one-line recipe broken up across multiple lines, which had no effect whatsoever.


hwine:
Re:
> I will remove the documentation in the master branch

Are you proposing that the source documents also be moved to the gh-pages branch?  Or could this be more precisely worded as "... remove the rendered documentation ..."?

.travis.yml
```diff
@@ -4,7 +4,16 @@ python:
...
 install:
-  - make install
+  - make install install-docs
 
 script:
-  - make black check_conftest_imports doctest coverage
+  - make black check_conftest_imports doctest coverage doc-build
+
+deploy:
+  provider: pages
```
hwine, line 13: Can you link to the documentation on this provider, please?  Maybe even as a comment for easy future reference.

smarnach:
I only plan to remove the rendered documentation. I'll update the PR description.

Makefile
```diff
@@ -65,12 +65,12 @@ black: check_venv
 install: venv
 	( . venv/bin/activate && pip install -U pip && pip install -r requirements.txt && python setup.py develop && pre-commit install )
 
+install-docs: venv
+	( . venv/bin/activate && pip install -r docs-requirements.txt )
```
hwine, line 69: I think the more common (modern) location for the requirements file needed for the docs is in the `docs/` directory. (Goes along with a separate Makefile in the docs directory.)  But this is certainly not a blocker.

Nit:: I believe the more common file name is `requirements-docs.txt` (4,074 on GitHub, 12 in Mozilla) `doc-requirements.txt` is less (1913, 0 in Mozilla), and `docs-requiements.txt` doesn't appear.

hwine (changes requested):
I'd like to review the travis provider docs before approving.

Makefile
smarnach: I moved the file to `docs/requirements.txt`. In the `docs` directory, I don't need any further qualification for the name.

smarnach:
@hwine Could you please take another look?

hwine (approved):
thanks! lgtm

smarnach:
@hwine Could you create a deploy key and add it Travis as GITHUB_TOKEN? I don't have the permissions to do so, and we should do that before merging this PR.

smarnach:
I deleted my last comment, since I saw the discussion on Slack.

PR #378, opened 2020-10-05
Add caching back to GCP

### LINKED ISSUE #363: Support in-run caching (for GCP)
Currently, when we use parameterize (and have no caching), we are making the same calls over and over if we run multiple tests that call for the same resources.

One way to solve this is to continue to use the current caching process, so that follow up calls to get certain resources just use the cache. (I believe this is how our caching works at the moment.)

The alternative is to utilize fixtures instead of parameterize. This will result in the fixture being cached during the pytest run.

In GCP, this only comes up as a problem when we are running a number of tests against a large number GCP projects.

ajvb:
fixes #363


g-k (approved):
r+ wfm

we might run into cache key length issues again if resource names get long

PR #380, opened 2020-10-06
Remove the generated docs.

smarnach:
They are [published via Travis now](https://github.com/mozilla/frost/pull/377).


hwine (approved):
could also delete `docs/index.html` and `docs/.nojekyll`

PR #382, opened 2020-10-09
Check EC2 EBS snapshots are not too old

bqbn:


aws/ec2/helpers.py
```diff
@@ -395,3 +395,24 @@ def ebs_volume_attached_to_instance(ebs, volume_created_days_ago=90):
...
+
+
+def ebs_snapshot_not_too_old(snapshot, snapshot_started_days_ago=365):
+    """
+    Check an ebs snapshot is created less than "snapshot_started_days_ago".
+
+    >>> from datetime import datetime
+    >>> from datetime import timezone
+    >>> from aws.ec2.helpers import ebs_snapshot_not_too_old
+    >>> ebs_snapshot_not_too_old({"StartTime": datetime.fromisoformat("2020-09-11T19:45:22.116+00:00")})
```
sciurus, line 407: This test will start failing a year from now. I think you want to use `datetime.now()` to get a timestamp less than 365 days old.

sciurus (changes requested):

aws/ec2/helpers.py
bqbn: That makes sense and is fixed.

ajvb (approved):

sciurus (approved):

PR #388, opened 2020-10-09
Add an rds check

bqbn:
Similar to https://github.com/mozilla/frost/pull/382, this is to check a RDS snapshot is not too old.

Also I probably won't submit more tests for a while since we have decided to shift efforts to enabling tests.

### DIFF
--- a/aws/rds/helpers.py
+++ b/aws/rds/helpers.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from helpers import get_param_id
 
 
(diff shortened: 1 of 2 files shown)


sciurus (changes requested):
@bqbn it looks like `rds_db_snapshot_not_too_old` and `ebs_snapshot_not_too_old` have the same logic. Can you refactor this and #382 to combine them into a single helper function that is shared between the rds and ebs snapshot tests?

I think the only real change is you'll need to move extracting the creation time into the test, and pass that rather than the snapshot into the helper.

bqbn:
@sciurus currently ec2 and rds have their own `helpers.py`. If we do what you requested, then in which file would you suggest putting the combined function? 

I think it's confusing if we end up doing `from aws.rds.helpers import rds_db_snapshot_not_too_old` in an ec2 module file, or `from aws.ec2.helpers import ebs_snapshot_not_too_old` in an rds module file.

sciurus:
Oh, that's a good point!

AJ is there some existing place to stick logic which we want tested that is useful across multiple resource types?

If not, then the duplication seems ok after all.

ajvb:
> Oh, that's a good point!
> 
> AJ is there some existing place to stick logic which we want tested that is useful across multiple resource types?
> 
> If not, then the duplication seems ok after all.

I don't believe we have a place defined atm, but `aws/helpers.py` would probably be the best place.

sciurus (approved):

ajvb (approved):

PR #391, opened 2020-10-16
Support for allowed ports for GCP firewall tests

ajvb:
Adds support for allowed ports in GCP firewall tests

As well, switch from the term "whitelisted" to "allowed".

Note: Will require a change to all of the configs in our automation code.

### DIFF
--- a/custom_config.py
+++ b/custom_config.py
@@ -29,6 +29,8 @@ class CustomConfig:
 class CustomConfigMixin:
     def __init__(self, config):
         self.user_is_inactive = config.get("user_is_inactive", {})
+        self.allowed_ports_global = set(config.get("allowed_ports_global", []))
+        self.allowed_ports = config.get("allowed_ports", [])
 
     def no_activity_since(self):
         no_activity_since = self._parse_user_is_inactive_relative_time(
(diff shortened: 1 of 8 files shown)


custom_config.py
```diff
@@ -54,13 +56,26 @@ def _parse_user_is_inactive_relative_time(self, key):
             weeks=+self.user_is_inactive[key].get("weeks", 0),
         )
 
+    def get_allowed_ports(self, test_id):
+        return self.get_allowed_ports_from_test_id(test_id) | self.allowed_ports_global
```
g-k, line 60: so the readme config allows port 25 to the bastion?

g-k (approved):
r+ lgtm but it would be good to document that test specific `allowed_ports` include `allowed_ports_global` too (it looks like that's the case)

custom_config.py
ajvb: yes

ajvb:
@g-k :+1: I added comments in the config file example and readme, also opened up https://github.com/mozilla/frost/issues/392 as I'd like to more completely document the config file

ajvb:
(going to wait till monday to merge so nothing breaks over the weekend)

PR #393, opened 2020-10-20
Alternate fix for GH-311 (ignore all tests)

hwine:
This reduces the scope of ignoring *.py files during test collection to runs via `frost`. When `pytest` is directly invoked, all normal pytest test selection methods are available.

aka "don't confuse pytest muscle memory" :D


ajvb (approved):
I'd say this makes sense. The hope is to move towards using `frost` as the command for nearly everything.

PR #399, opened 2020-10-23
docs: development: add section on running

aerickson:


hwine (changes requested):
For development, you definitely need to clone the repo, and not install via pip. Since pip install is currently broken (#401), the "I just want to use Frost" instruction are (for now) what the development instructions should be. See the [installation section](https://github.com/mozilla/frost/blob/master/README.md#installing for the terse version.

ajvb:
Thanks @aerickson! We added https://github.com/mozilla/frost/pull/400 around the time this was originally opened. We'll keep with that one till we can confidently go back to recommending `pip install frost` or similar.

Issue #385, opened 2020-10-09
Roll service_report_generator.py into the frost CLI


ajvb:
Now that we have a frost CLI it might be nice to roll https://github.com/mozilla/frost/blob/master/service_report_generator.py into it.

ajvb:
Alongside this, we should add documentation around the JSON output that `server_report_generator.py` creates, beyond what is documented in the script itself.

hwine:
So this is an interesting architectural decision for me with GitHub.  I don't really understand the script, so was planning on consuming the JSON directly from frost in my post-test-actions.

What are the reasons I should reconsider that decision?

PR #403, opened 2020-10-26
Fix exemptions

ajvb:
Have exemptions work as expected by properly parsing out the "test id"
(defined by the parameterize's `id` func)


ajvb:
cc: @sciurus - this is why the original piops exemptions were not working :(

exemptions.py
```diff
@@ -125,11 +125,26 @@ def add_xfail_marker(item):
...
 
     test_exemptions = item.config.custom_config.exemptions.get(item.originalname, None)
-    # item is a _pytest.python.Function
-    # e.g. gsuite/admin/test_admin_user_is_inactive.py::test_admin_user_is_inactive[gsuiteuser@example.com]
-    test_id = item.nodeid
+
+    test_id = item.name
+    try:
+        # test_admin_user_is_inactive[gsuiteuser@example.com]
+        test_id = item.name.split("[")[1][:-1]
```
hwine, line 132: Nit: the string you're parsing is actually `{relative_path_to_file}::{test_name}[{parameters}]`. Any reason you're not parsing `metadata.parametrized_name` (which is in the format of the comment).

Related (but not for this PR), I also wonder if we should make "parametrized" a separate field in the JSON output (first level `metadata`). There is already `parametrized_name` (`test_pho[params]`) and `unparametrized_name` (just `test_pho`), so we have a bit of a precedent.

hwine (approved):

exemptions.py
ajvb: `item.name` is the same as `metadata.parametrized_name` - https://github.com/mozilla/frost/blob/master/conftest.py#L302

I decided it might be easier to repeat the use of `item.name` here rather than combine this with `metadata.parametrized_name` since this hook is called earlier and the `pytest_runtest_makereport` is operating on the report rather than the item.

hwine: Interesting -- in the JSON output file, the two values (`name` & `metadata.parametrized_name`) are different -- they must get munged later.

Is it worth documenting the call sequencing? Or will this just serve as "thar be dragons!" to keep mere mortals awsy.

ajvb: I'd say it's worth documenting, as a part of https://github.com/mozilla/frost/issues/385

`service_report_generator.py` has been outside of "Frost core" so far, and imo this needs to be resolved in conjunction with better documentation around it.

PR #404, opened 2020-10-27
[Doc] explain parametrization conventions

hwine:
first draft -- please check for correctness.  [rendered view](https://hwine.github.io/frost/NewServices.html#add-service-specific-content)


hwine:
Note: I refer to `METADATA_KEYS` as a set. Currently it is a list, but it should be safe to update that. If there are no objections, I'll prepare a bug & PR.

ajvb (approved):

PR #405, opened 2020-10-29
Change METADATA_KEYS to a set

hwine:
Report (by exception) if duplicate values are added, as that is almost
certainly not intended.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -196,7 +196,7 @@ class SingleSet(set):
     SingleSet({1, 2, 3, 4, 5, 6})
     >>>
 
-    **BUG:**
+    **NB:**
     - duplicate values on initialization are not detected
     >>> ss = SingleSet({1, 2, 3, 4, 3, 2, 1})
     >>> ss
--- a/conftest.py
+++ b/conftest.py
@@ -165,30 +165,89 @@ def get_node_markers(node):
     return [m for m in node.iter_markers()]
 
 
-METADATA_KEYS = [
-    "DBInstanceArn",
-    "DBInstanceIdentifier",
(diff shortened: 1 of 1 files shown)


conftest.py
```diff
@@ -165,30 +165,89 @@ def get_node_markers(node):
...
+    >>> ss.update({4, 5, 6, 3})
+    Traceback (most recent call last):
+    ...
+    conftest.DuplicateKeyError: Value(s) {3, 4} already present
+    >>> ss
+    SingleSet({1, 2, 3, 4, 5, 6})
+    >>>
+
+    **BUG:**
+    - duplicate values on initialization are not detected
```
ajvb, line 200: Is this a bug? It's interesting, cause I could see this just an understandable inheritance of the normal Python set. But if we are considering this a bug we should fix it. If not, then let's change this to just be a description of it's behavior?

conftest.py
```diff
@@ -165,30 +165,89 @@ def get_node_markers(node):
...
+    conftest.DuplicateKeyError: Value 3 already present
+    >>> ss.update({4, 5, 6, 3})
+    Traceback (most recent call last):
+    ...
+    conftest.DuplicateKeyError: Value(s) {3, 4} already present
+    >>> ss
+    SingleSet({1, 2, 3, 4, 5, 6})
+    >>>
+
+    **BUG:**
```
hwine, line 199:
```suggestion
    **NB:**
```

conftest.py
hwine: Good point -- making it just an NB

ajvb (approved):

PR #407, opened 2020-11-02
Fix Frost exit codes

ajvb:
and some little additional fixes

CI has been broken for doctests since the switch to using `frost test`


hwine (approved):
Thanks for fixing these -- this may improve the doc rendering, too.

PR #413, opened 2020-11-11
Update cache

g-k:
Should make it easier to land caching changes for the next milestone with breaking changes.
 
`mypy --strict cache.py` passes.


ajvb (approved):
Woah this is awesome!

Issue #355, opened 2020-09-21
add type annotations


g-k:
e.g. with dataclass or typeddict from the stdlib or if botocore provides types. 

This would prevent issues and make it clearer what's being passed around.

We can try to infer types with https://monkeytype.readthedocs.io/en/latest/generation.html at a quick pass I didn't see an obvious way to invoke the frost CLI.

hwine:
This ticket covers ways to update legacy code. For all new code, use of type hints is encouraged.

PR #414, opened 2020-11-11
add type annotations to aws/{client,conftest}.py

g-k:
Type annotations should make it easier to make changes or detect breaking changes.

Might need to tweak or tighten some of the annotations from Any.

refs: #355

### DIFF
--- a/aws/client.py
+++ b/aws/client.py
@@ -4,18 +4,29 @@ import os
 from collections import namedtuple
 import functools
 import itertools
+from typing import (
+    Any,
+    Callable,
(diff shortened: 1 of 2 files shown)


ajvb (approved):
looks awesome!

As we make changes like these, should we add a `mypy --strict` check within CI? As well, within precommit?

g-k:
> As we make changes like these, should we add a mypy --strict check within CI? As well, within precommit?

Yeah, we'll probably need to build out a config to ignore missing imports and calls to untyped files.

PR #416, opened 2020-11-12
Cache filename tweaks

### LINKED ISSUE #105: Cache key is not unique when --aws-profiles is not used
The [`cache_key`](https://github.com/mozilla-services/pytest-services/blob/master/aws/client.py#L73-L93) function uses the passed in `profile` as part of its cache key. As a result, when you are using env vars instead of `--aws-profiles`, you can have cached results be used when they are from an entirely different account.

### LINKED ISSUE #349: filename too long breaking the cache
Some OSes have filename limits. @Micheletto ran into a "filename to long" error for an SNS topic call resulting in the AWS client producing `cache_key` filename longer than 255 bytes.

ajvb:
* Use directory struture for cache keys / filenames - Related to #349
* Use hashlib.md5 to hash args part of cache key for AWS - Fixes #349
* Get and use AWS account id if profile is None - Fixes #105

### DIFF
--- a/aws/client.py
+++ b/aws/client.py
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
 
-import os
-from collections import namedtuple
 import functools
 import itertools
+import os
+import warnings
+from collections import namedtuple
+from hashlib import md5
 from typing import (
     Any,
     Callable,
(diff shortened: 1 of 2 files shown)


g-k (approved):
r+ lgtm

PR #417, opened 2020-11-16
Make sure all test id's are reproducible

### LINKED ISSUE #412: Make sure all test id's are reproducible
Currently, there are a number of tests where part of the test id / resource id is random. This is usually due to the `@pytest.mark.parametrize` call having two resources but the id function only support one of them, i.e. https://github.com/mozilla/frost/blob/master/aws/iam/test_iam_user_without_mfa.py#L8-L12

This leads to parametrize id's looking something like this: `test_iam_user_without_mfa[ajvb-iam_user_mfa_devices9]`

In this case, the number at the end of `iam_user_mfa_devices9` is based on the order in which the test ran, meaning it is likely to change in the next run.

At the moment, this is not breaking anything, but it is a blocker for having state management around alerts / bugs created from results. Converting that test name and resource id into something key-able would require finicky custom logic.

Instead, all parametrize id's should be reproducible, so having it as `test_iam_user_without_mfa[ajvb]` or `test_iam_user_without_mfa[ajvb-<constant-device-id>]`

ajvb:
Fixes the `ids` argument in all `parameterize` calls that make use of `zip()` so that the outputted id is actually generated.

Fixes #412

### DIFF
--- a/aws/s3/helpers.py
+++ b/aws/s3/helpers.py
@@ -5,7 +5,31 @@ def get_s3_bucket_name(bucket):
     return get_param_id(bucket, "Name")
 
 
-def get_s3_bucket_name_only(bucket):
-    if isinstance(bucket, dict) and "Name" in bucket:
-        return get_s3_bucket_name(bucket)
+def get_s3_resource_id(resource):
+    if isinstance(resource, dict) and "Name" in resource:
+        return get_s3_bucket_name(resource)
+    if isinstance(resource, dict) and "ID" in resource:
+        return get_param_id(resource, "ID")
+    if isinstance(resource, dict) and "Owner" in resource:  # ACL
+        return get_param_id(resource["Owner"], "DisplayName")
+    if isinstance(resource, dict) and "Status" in resource:  # Versioning
+        return get_param_id(resource, "Status")
+    if isinstance(resource, dict) and "AllowedHeaders" in resource:  # CORS
+        return "cors-rules"
+
+    if isinstance(resource, dict) and "ResponseMetadata" in resource:
+        return "empty"
+
+    if isinstance(resource, dict) and not resource:
+        return "empty"
+
+    if isinstance(resource, list):
+        if len(resource) == 0:
+            return "empty"
+        else:
+            return get_s3_resource_id(resource[0])
+
+    if resource is None:
+        return "none"
+
     return None
--- a/aws/iam/helpers.py
+++ b/aws/iam/helpers.py
@@ -152,7 +152,10 @@ def get_iam_user_name(login):
     return get_param_id(login, "UserName")
 
 
-def get_iam_user_name_only(user):
-    if isinstance(user, dict) and "UserName" in user:
-        return get_iam_user_name(user)
+def get_iam_resource_id(resource):
+    if isinstance(resource, dict) and "UserName" in resource:
+        return get_iam_user_name(resource)
+    if isinstance(resource, list):
+        if len(resource) == 0:
+            return "empty"
     return None
--- a/aws/iam/resources.py
+++ b/aws/iam/resources.py
@@ -132,18 +132,18 @@ def iam_admin_login_profiles():
 
 
 def iam_admin_mfa_devices():
-    "botocore.readthedocs.io/en/latest/reference/services/iam.html#IAM.Client.list_mfa_devices"
+    "https://botocore.readthedocs.io/en/latest/reference/services/iam.html#IAM.Client.list_mfa_devices"
     return iam_mfa_devices(iam_admin_users())
 
 
 def iam_user_login_profiles():
     "http://botocore.readthedocs.io/en/latest/reference/services/iam.html#IAM.Client.get_login_profile"
-    return iam_login_profiles([user for user in iam_users()])
+    return iam_login_profiles(iam_users())
 
 
 def iam_user_mfa_devices():
-    "botocore.readthedocs.io/en/latest/reference/services/iam.html#IAM.Client.list_mfa_devices"
-    return iam_mfa_devices([user for user in iam_users()])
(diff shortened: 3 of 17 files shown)


g-k (approved):
r+ lgtm

Not sure how hard it was to make all these changes, but since we're using tests as code maybe we should look into tools for editing the concrete syntax tree (e.g. something based on https://pypi.org/project/libcst/ or another library I couldn't dig up).

aws/s3/helpers.py
```diff
@@ -5,7 +5,31 @@ def get_s3_bucket_name(bucket):
     return get_param_id(bucket, "Name")
 
 
-def get_s3_bucket_name_only(bucket):
-    if isinstance(bucket, dict) and "Name" in bucket:
-        return get_s3_bucket_name(bucket)
+def get_s3_resource_id(resource):
+    if isinstance(resource, dict) and "Name" in resource:
+        return get_s3_bucket_name(resource)
```
hwine, line 10: This structure feels wrong to me -- shouldn't it return a concatenation of all the matching key's values?

I.e. is it guaranteed (or a requirement) that "resource" have at most one of the keys of interest?  If so, I'd love to see that as a comment.

hwine (changes requested):
Looking for an answer and/or a comment

hwine:
> 
> Not sure how hard it was to make all these changes, but since we're using tests as code maybe we should look into tools for editing the concrete syntax tree (e.g. something based on https://pypi.org/project/libcst/ or another library I couldn't dig up).

You lost me -- what's the maintenance issue you see (i.e. use case), and how would libcst solve that?

aws/s3/helpers.py
ajvb: No, as the test id serves two functions:
* A unique identifier for the resource(s) being tested
* A best effort name for the resource(s) in question

For certain tests, the test id will work as all of the additional data you need to know, i.e. the test id for MFA tests being the username. But for a number of other tests, i.e. security group / firewall tests, it's just there to point you to the security group in question which you will then have to go and look at via the AWS/GCP tooling. Or is some other cases, you will just need to look at the metadata associated with the test result.

If a resource has multiple matching keys, the ordering is in place to attempt to get the most representative identifier for the resource.

hwine: Ahh, if there is a "magical order" to the statements, I _really_ think that needs to be documented in the code. Otherwise it's extremely brittle for any maintainer who doesn't grok the full service ontology.

Especially since I could see this code being an "obvious" modification point for the "file name too long" issue.

ajvb: There isn't a magic order, but nevertheless I do agree that this current way of handling test ids is extremely brittle.

I'm beginning to think that it might make sense to bring these "get resource id" functions closer to the calls within `resources.py`. Maybe the current `resources.py` functions should be turned into classes with a "get_resources" and a "get_id" function required for each, with some mixins that can be used for common ids. Would additionally require a refactor around how we use `parameterize`, but I think that would be worth it.

g-k:
> > Not sure how hard it was to make all these changes, but since we're using tests as code maybe we should look into tools for editing the concrete syntax tree (e.g. something based on https://pypi.org/project/libcst/ or another library I couldn't dig up).
> 
> You lost me -- what's the maintenance issue you see (i.e. use case), and how would libcst solve that?

Safely editing a bunch of test and resource files e.g. converting all `str.format` calls to f-strings https://libcst.readthedocs.io/en/latest/codemods_tutorial.html

aws/s3/helpers.py
g-k: > I'm beginning to think that it might make sense to bring these "get resource id" functions closer to the calls within resources.py. Maybe the current resources.py functions should be turned into classes with a "get_resources" and a "get_id" function required for each, with some mixins that can be used for common ids. Would additionally require a refactor around how we use parameterize, but I think that would be worth it.

+1 dataclasses or a validator library could help too.

At some point I think it makes sense to switch to boto3 since it looks like it provides and ID function https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/resources.html#boto3.resources.model.Identifier (it'd also be more familiar to Ops and provide more pythonic attribute access)

hwine: fwiw - see also my approach at:
 - [usage](https://github.com/hwine/frost/blob/branch_check/github/branches/test_branch_protection.py#L27-L29)
- definition [in dataclass](https://github.com/hwine/frost/blob/branch_check/github/branches/validate_compliance.py#L19-L24)

I'm not sure a mixin is justified -- we're simply adding one method. Mixins come with complexity and future maintenance issues.

ajvb: I've combined this discussion into a new issue: https://github.com/mozilla/frost/issues/418

@hwine Are you ok with us merging this as-is and keeping the above discussion as a follow-up issue?

hwine:
> > You lost me -- what's the maintenance issue you see (i.e. use case), and how would libcst solve that?
> 
> Safely editing a bunch of test and resource files e.g. converting all `str.format` calls to f-strings https://libcst.readthedocs.io/en/latest/codemods_tutorial.html

Ah, there's a `pre-commit` plugin for that (and more): [pyupgrade](https://github.com/asottile/pyupgrade). Besides fstrings, it also catches other python changes that the code can take advantages of.  See the [README](https://github.com/asottile/pyupgrade)

I'd be happy to see that plugin added to frost -- it'll help when we have to go to python 3.9 (yearly updates now). I think all of it's changes should be a separate PR for a "flag day", so we can all do our merges-for-active-branches prior, and have that change apply cleanly.

aws/s3/helpers.py
hwine: go for it

hwine (approved):
Thanks for opening the related issue -- we're good to go

PR #425, opened 2020-12-04
Fixes COPY so that frost works as expected

### LINKED ISSUE #397: Running frost from docker image fails
On master, if I run

```
$ make build-image
$ docker run localhost/frost:latest test aws/ec2/test_ec2_ebs_volume_not_piops.py 
```

I get the error

```
ERROR: usage: frost [options] [file_or_dir] [file_or_dir] [...]
frost: error: unrecognized arguments: --debug-calls
  inifile: None
  rootdir: /app
```

Based on some experimentation, I think this is because the Dockerfile runs `python setup.py install` but the frost command currently only works if you run `python setup.py develop`. Here are the steps to reproduce the problem outside of Docker:

1) Perform a fresh checkout of the frost repo
2) Run `. venv/bin/activate && pip install -U pip && pip install -r requirements.txt && python setup.py install && pre-commit install` (the standard `make install` steps, except with _install_ substituted for _develop_)
3) Run `frost test aws/ec2/test_ec2_ebs_volume_not_piops.py `
4) Observe the error "unrecognized arguments: --debug-calls"

So we need to either figure out why frost doesn't work when you run `install` or switch the Dockerfile to run `develop`.

ajvb:
the original `COPY * ./` was actually copying sub-directories rather than the contents of the directory, so we want `COPY . /app`

Fixes #397

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,11 +18,9 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
             git \
             jq
 
+COPY . /app
 WORKDIR /app
 
-
-COPY frost/* frost/
-COPY * ./
 RUN python setup.py install
 
 USER app


hwine (approved):
missing documentation: "as expected" is not defined anywhere -- #397 lists several options.

It would be good to know which options this supports, so we update use cases, especially for the cases we don't support.

ajvb:
Good point. I was defining "as expected" as what is documented in the "Running" section of the README (https://github.com/mozilla/frost#running), but would be good to specify that for Docker within the README.

PR #428, opened 2020-12-10
Add test for subscriptions without topics.

Micheletto:
This PR intends to do two things:

1. Stop the `sns_subscription_attributes()` function from crashing the **test_sns_pending_verified.py** test when the AWS API throws an error on getting attributes for a subscription that does not have a parent topic.
1. Add a test to find subscriptions that have no parent topic.

It works as is, but there are few things that need fixing before a merge.

- The method used to truncate an SNS subscription ARN into an SNS topic ARN is not great:
```
topic = ':'.join(re.split(':', s_arn)[0:6])
```
- Several functions added might be moved into an SNS **helpers.py** file instead of their current location in **resources.py**.
  - `sns_parent_topic_exists()`
  - `sns_get_subscription_attrs()`

No doubt there are other pythonic issues.

### DIFF
--- a/aws/sns/resources.py
+++ b/aws/sns/resources.py
@@ -1,5 +1,4 @@
 from conftest import botocore_client
-import re
 
 
 def sns_subscriptions():
@@ -22,21 +21,12 @@ def sns_topics():
     )
 
 
-def sns_parent_topic_exists(subscription):
-    "https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#subscription"
-    topicArn = ":".join(
-        re.split(":", subscription)[0:6]
-    )  # Naive method of divining topic from subscription
-    try:
-        botocore_client.get(
-            service_name="sns",
-            method_name="get_topic_attributes",
-            call_args=[],
-            call_kwargs={"TopicArn": topicArn},
-        )
-        return True
-    except:
-        return False
+def sns_topic_arns():
+    "https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#topic"
+    topicArns = {}
+    for x, y in enumerate(sns_topics()):
+        topicArns[y["TopicArn"]] = 1
+    return topicArns
 
 
 def sns_get_subscription_attrs(subscription):
--- a/aws/sns/resources.py
+++ b/aws/sns/resources.py
@@ -1,4 +1,5 @@
 from conftest import botocore_client
+import re
 
 
 def sns_subscriptions():
@@ -19,6 +20,23 @@ def sns_topics():
         .flatten()
         .values()
     )
+
+
+def sns_parent_topic_exists(subscription):
+    "https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#subscription"
+    topicArn = ":".join(
+        re.split(":", subscription)[0:6]
+    )  # Naive method of divining topic from subscription
+    try:
+        botocore_client.get(
+            service_name="sns",
+            method_name="get_topic_attributes",
+            call_args=[],
+            call_kwargs={"TopicArn": topicArn},
+        )
+        return True
+    except:
+        return False
 
 
 def sns_get_subscription_attrs(subscription):
--- a/aws/sns/resources.py
+++ b/aws/sns/resources.py
@@ -1,4 +1,6 @@
+import botocore
 from conftest import botocore_client
+from botocore.errorfactory import ClientError
 
 
 def sns_subscriptions():
@@ -47,8 +49,11 @@ def sns_get_subscription_attrs(subscription):
             .values()[0]
         )
         return attrs
-    except:
-        pass
+    except botocore.errorfactory.ClientError as e:
+        if e.response["Error"]["Code"] != "InvalidParameter":
+            raise
+    # except:
+    #    pass
 
 
 def sns_subscription_attributes():
--- a/aws/sns/resources.py
+++ b/aws/sns/resources.py
@@ -1,6 +1,4 @@
-import botocore
 from conftest import botocore_client
-from botocore.errorfactory import ClientError
 
 
 def sns_subscriptions():
@@ -25,42 +23,7 @@ def sns_topics():
 
 def sns_topic_arns():
     "https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#topic"
-    topicArns = {}
-    for x, y in enumerate(sns_topics()):
-        topicArns[y["TopicArn"]] = 1
-    return topicArns
-
-
-def sns_get_subscription_attrs(subscription):
-    "https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#subscription"
-    # This function catches the error thrown by the AWS API when attempting to query the attributes of
-    # a subscription that has no parent topic.
-    try:
-        attrs = (
-            botocore_client.get(
-                service_name="sns",
-                method_name="get_subscription_attributes",
-                call_args=[],
-                call_kwargs={"SubscriptionArn": subscription["SubscriptionArn"]},
-                profiles=[subscription["__pytest_meta"]["profile"]],
-                regions=[subscription["__pytest_meta"]["region"]],
-            )
-            .extract_key("Attributes")
-            .values()[0]
-        )
-        return attrs
-    except botocore.errorfactory.ClientError as e:
-        if e.response["Error"]["Code"] != "InvalidParameter":
-            raise
-    # except:
-    #    pass
-
-
-def sns_subscription_attributes():
-    "https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#subscription"
-    return [
-        sns_get_subscription_attrs(subscription) for subscription in sns_subscriptions()
-    ]
+    return {x["TopicArn"] for x in sns_topics()}
 
 
 def sns_subscriptions_by_topic():
--- a/aws/sns/resources.py
+++ b/aws/sns/resources.py
@@ -21,20 +21,32 @@ def sns_topics():
     )
 
 
+def sns_get_subscription_attrs(subscription):
+    "https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#subscription"
+    # This function catches the error thrown by the AWS API when attempting to query the attributes of
+    # a subscription that has no parent topic.
+    try:
+        attrs = (
+            botocore_client.get(
+                service_name="sns",
+                method_name="get_subscription_attributes",
+                call_args=[],
+                call_kwargs={"SubscriptionArn": subscription["SubscriptionArn"]},
+                profiles=[subscription["__pytest_meta"]["profile"]],
+                regions=[subscription["__pytest_meta"]["region"]],
+            )
+            .extract_key("Attributes")
+            .values()[0]
+        )
+        return attrs
+    except:
+        pass
+
+
 def sns_subscription_attributes():
     "https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#subscription"
     return [
-        botocore_client.get(
-            service_name="sns",
-            method_name="get_subscription_attributes",
-            call_args=[],
-            call_kwargs={"SubscriptionArn": subscription["SubscriptionArn"]},
-            profiles=[subscription["__pytest_meta"]["profile"]],
-            regions=[subscription["__pytest_meta"]["region"]],
-        )
-        .extract_key("Attributes")
-        .values()[0]
-        for subscription in sns_subscriptions()
+        sns_get_subscription_attrs(subscription) for subscription in sns_subscriptions()
     ]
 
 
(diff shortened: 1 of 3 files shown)


aws/sns/resources.py
```diff
@@ -21,20 +22,49 @@ def sns_topics():
     )
 
 
-def sns_subscription_attributes():
+def sns_parent_topic_exists(subscription):
     "https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#subscription"
-    return [
+    topicArn = ":".join(
+        re.split(":", subscription)[0:6]
+    )  # Naive method of divining topic from subscription
```
smarnach, line 29: I'm not sure the format is specified and official part of the interface, but I wouldn't really expect this to change, so what you have seems fine. You could use `subscription.split(":")` (using the `str` method `split()`, which splits on literal strings rather than regular expressions).

If I run `aws sns list-subscriptions` from the command line, I get a list that includes both the subscription ARN and the topic ARN for each subscription. Maybe we can pass the whole JSON object (i.e. Python dict) to this function, rather than just the subscription ARN?

aws/sns/resources.py
```diff
@@ -21,20 +22,49 @@ def sns_topics():
...
-            method_name="get_subscription_attributes",
+            method_name="get_topic_attributes",
             call_args=[],
-            call_kwargs={"SubscriptionArn": subscription["SubscriptionArn"]},
-            profiles=[subscription["__pytest_meta"]["profile"]],
-            regions=[subscription["__pytest_meta"]["region"]],
+            call_kwargs={"TopicArn": topicArn},
+        )
+        return True
+    except:
```
smarnach, line 38: It could be helpful to be a bit more specific what errors to catch – catching, e.g., `TypeError` here could make for some fun debugging.

aws/sns/resources.py
```diff
@@ -21,20 +22,49 @@ def sns_topics():
...
+    # a subscription that has no parent topic.
+    try:
+        attrs = (
+            botocore_client.get(
+                service_name="sns",
+                method_name="get_subscription_attributes",
+                call_args=[],
+                call_kwargs={"SubscriptionArn": subscription["SubscriptionArn"]},
+                profiles=[subscription["__pytest_meta"]["profile"]],
+                regions=[subscription["__pytest_meta"]["region"]],
```
smarnach, line 54: I still have an unfinished PR to get rid of these ugly parameters, but didn't have time to finish it up yet.

aws/sns/resources.py
```diff
@@ -21,20 +22,49 @@ def sns_topics():
...
-        for subscription in sns_subscriptions()
+        return attrs
+    except:
+        pass
+
+
+def sns_subscription_attributes():
+    "https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns.html#subscription"
+    return [
+        sns_get_subscription_attrs(subscription) for subscription in sns_subscriptions()
```
smarnach, line 67: The function `sns_get_subscription_attrs()` defined above will return `None` whenever it encounters and ignores an error, so there would be random `None` values in this list, which I don't think is the intended behaviour.

aws/sns/resources.py
```diff
@@ -21,20 +22,49 @@ def sns_topics():
...
         botocore_client.get(
             service_name="sns",
-            method_name="get_subscription_attributes",
+            method_name="get_topic_attributes",
             call_args=[],
-            call_kwargs={"SubscriptionArn": subscription["SubscriptionArn"]},
-            profiles=[subscription["__pytest_meta"]["profile"]],
-            regions=[subscription["__pytest_meta"]["region"]],
+            call_kwargs={"TopicArn": topicArn},
+        )
```
smarnach, line 36: We have some kind of caching of API results, right? So maybe it would be nicer to call `list_topics` here, and check whether the topic is in the list. If the result of `list_topics` is indeed cached, we'd only make a single API call for all subscriptions, instead of one for each subscription.

aws/sns/resources.py
```diff
@@ -21,20 +22,49 @@ def sns_topics():
...
+                regions=[subscription["__pytest_meta"]["region"]],
+            )
+            .extract_key("Attributes")
+            .values()[0]
         )
-        .extract_key("Attributes")
-        .values()[0]
-        for subscription in sns_subscriptions()
+        return attrs
+    except:
```
smarnach, line 60: A more specific `except` clause would also be nice here.

aws/sns/resources.py
Micheletto: My objection was primarily to using a split, slice, join operation to lop the end off a string.  Using a string method instead of a a regular expression is certainly an improvement.  I wasn't aware that the `list-subscriptions` returned the topic ARN.  That's certainly the way to go here.

aws/sns/resources.py
Micheletto: I had thought about doing that originally, and not being pythonically inclined, I couldn't think of how to go about doing it.  However, as you suggest, it really is a better solution and I _think_ I can mange it.

aws/sns/resources.py
Micheletto: I'm going to take a stab at checking the topics list instead so catching this error should no longer be required.  If we have to go this route I'll add TypeError here.

aws/sns/resources.py
Micheletto: Cool.

aws/sns/resources.py
Micheletto: Do you have any suggestions on what to add?  I'm a bit out of my depth here.

aws/sns/resources.py
Micheletto: I'm not certain what the best thing to return here would be.  It's unable to check the PendingVerification status, so maybe we still want these to fail.

aws/sns/resources.py
smarnach: If the API call is implicitly cached, you don't need to change much – just call `list_topics` here every time this is called, and the list will be retrived from the cache. However, I'm not sure how that weird cache works, so you'd need to figure out whether it's actually cached.

aws/sns/resources.py
smarnach: There's a specific error you want to avoid here, right? The current code basically says "whatever goes wrong, just paper over it and carry on." What I believe you actually want is "I know the API call above sometimes throws this specific error, and you don't need to mind that one."

So if you can reproduce this error, or have any execution log of that error occurring, you can see what the error class of that error was `botocore.exceptions.ClientError`, then you can import that error class into this module and do

    except botocore.exceptions.ClientError:

here. I don't know what the right error class is – that depends on the errors you have encountered here.

This has two advantages over the current code. First, it makes it clearer what error you want to ignore, which helps readers of the code. (Code is read much more often than it is written.) And second, it doesn't paper over some random other errors you didn't know could occur here, so you will still see unexpected errors in the logs.

In some cases you may want to be even more specific than restricing the errors you catch to a specific class. In these cases you can inspect the exception object an re-raise it isn't what you expected:

    except GlonkError as exc:
        if exc.code != 2345:        # completely made up – custom exceptions
                                    # usually have fields with additional info
            raise                   # reraise exception if it doesn't have the right code

See also [this StackOverflow question](https://stackoverflow.com/questions/46174385/properly-catch-boto3-errors) for a more on-topic example.

Or you can be less specific, by catching multiple classes with `except TypeError, ValueError:` or base classes of exceptions, which will match all exceptions derived from that base class. Unless you have specific requirements, I'd keep it simple at first and just catch the exception by class.

You almost never want a bare `except:` class without any further restriction. This will also catch `KeyboardInterrupt` and `SystemExit` and just pretend it didn't happen, which probably isn't what you meant. To broadly catch all application exceptions, you can use `except Exception:`; `Exception` is the base class of most built-in exceptions and virtually all user-defined exceptions, so this is almost always better than just `except:`.

(Sorry, this was probably more than you asked for – I somehow switched to StackOverflow mode. :-) )

aws/sns/resources.py
smarnach: It depends on what you want. Currently, `sns_get_subscription_attrs()` will return `None` on error, which may be reasonable behaviour. If it is, you probably want to filter the None values out from the list before returnning it, e.g. with `[x for x in my_list if x is not None]`.

Or maybe it's an actual error when you can't retrieve details for one of the subscriptions? In that case, don't catch it, or at least don't catch it in `sns_get_subscription()`, but further up the call stack, where you know what to do with it.

Or maybe you can still get details for these subscriptions by asking the API in a different way? I just don't know, since I don't understand what that error condition ("unable to check the PendingVerification status") means.

aws/sns/resources.py
smarnach: Agreed, if possible it's best to use the topic ARN returned by the API.

aws/sns/resources.py
ajvb: From my understanding, we'd probably want to go with the `[x for x in my_list if x is not None]` method.

aws/sns/resources.py
Micheletto: It may have been more than I _asked_ for, but it was exactly what I needed.  Thank you for the very detailed response.  I believe this latest commit sufficiently limits the exception handling to the specific error condition we want to trap.

Micheletto:
I believe this is ready to merge.

Some comments for the PR: 
>     1. Stop the `sns_subscription_attributes()` function from crashing the **test_sns_pending_verified.py** test when the AWS API throws an error on getting attributes for a subscription that does not have a parent topic.
No longer querying subscription attributes.


> ```
> topic = ':'.join(re.split(':', s_arn)[0:6])
> ```
No longer doing this.


>       * `sns_parent_topic_exists()`
>       * `sns_get_subscription_attrs()`

These functions are gone.

aws/sns/test_sns_pending_verified.py
```diff
@@ -1,14 +1,21 @@
...
-    ids=lambda subscription: get_param_id(subscription, "SubscriptionArn"),
+    "topic",
+    sns_subscriptions_by_topic(),
+    ids=lambda topic: get_param_id(topic, "TopicArn"),
 )
-def test_sns_pending_verified(subscription_attrs):
-    assert subscription_attrs["PendingConfirmation"] == "false"
+def test_sns_topics_without_subscriptions(topic):
+    for subscription in topic["Subscriptions"]:
+        assert subscription["SubscriptionArn"] != "PendingConfirmation"
```
smarnach, line 21: Something looks off here. Will the subscription ARN really be "PendingConfirmation"?

smarnach (commented):
Nice, this got a lot simpler! There is still one test I don't fully understand.

aws/sns/test_sns_pending_verified.py
Micheletto: Yeah, discovered that in testing.  I cannot find another way to do this.  It appears that AWS does not assign an ARN until a subscription is no longer in the PendingConfirmation state.

ajvb: My concern with this test is wondering whether it will fail frequently on newly created subscriptions? Are there cases where a subscription might go unconfirmed for a day or two? If that is the case, can we add a note about that in the rationale text?

ajvb (approved):

aws/sns/test_sns_pending_verified.py
smarnach: Or maybe we can change the test to only complain about unconfirmed subscriptions that are older than, say, a week?

Micheletto:
>My concern with this test is wondering whether it will fail frequently on newly created subscriptions? Are there cases where a subscription might go unconfirmed for a day or two? If that is the case, can we add a note about that in the rationale text?

I don't think it will.  For starters, we don't actually have any SNS subscriptions that are _pending confirmation_.  It's conceivable that someone would create one and fail to acknowledge it in a timely fashion, but in my opinion that's an error and should be reported as such: because resources that exist, but are unusable are problematic.

aws/sns/test_sns_pending_verified.py
Micheletto: My concern with this test is wondering whether it will fail frequently on newly created subscriptions? Are there cases where a subscription might go unconfirmed for a day or two? If that is the case, can we add a note about that in the rationale text?

I don't think it will. For starters, we don't actually have any SNS subscriptions that are pending confirmation. It's conceivable that someone would create one and fail to acknowledge it in a timely fashion, but in my opinion that's an error and should be reported as such: because resources that exist, but are unusable are problematic.

As far as waiting a week, I think that delay should exist in our automation not in the test.  I also, don't think it's currently a problem we need to solve, since we don't presently have any subscriptions pending confirmation.

Micheletto: I am not against adding something about it in the rationale, but I'm not sure what to put beyond what's there already.

ajvb: Yeah, I was going to recommend the "only fail after X hours", but I didn't see a created at field for subscriptions :shrug: 

>It's conceivable that someone would create one and fail to acknowledge it in a timely fashion, but in my opinion that's an error and should be reported as such: because resources that exist, but are unusable are problematic.

Ok yeah, I agree. Sounds good to me to leave as-is.

smarnach (approved):

PR #432, opened 2021-01-08
add GHA CI

g-k:
refs: #429

Migrate CI to github actions

NB: test w/ python 3.9 fails due to a dep build failure, but I think that's fine for now and we can not require it to pass to merge

NB: there are third party actions for publish GH pages we could try too (pending GH admin approval)

I don't have a way to test the publish actions. I'm least certain about the `gh-pages` action in particular.

- [ ] generate and add new pypi creds to the workflow config
- [ ] generate and add a new GH token with repo scope to the workflow config
- [ ] update the required jobs in branch protection settings
- [ ] tag a commit to test the publish workflow
- [ ] remove the travis CI config; update badges (separate PR)

r? @ajvb @hwine

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,45 @@
+name: Test
+
+on: [push, pull_request]
+
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.8]
+        os: [ubuntu-latest, ubuntu-20.04]
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Get pip cache dir
+        id: pip-cache
+        run: |
+          echo "::set-output name=dir::$(pip cache dir)"
+
+      - name: Cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.pip-cache.outputs.dir }}
+          key:
+            ${{ matrix.os }}-${{ matrix.python-version }}-${{ hashFiles('**/setup.py') }}
+          restore-keys: |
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,45 @@
...
+on: [push, pull_request]
+
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.8, 3.9]
```
ajvb, line 12: Should we disable `3.9` for now?

ajvb (approved):
Awesome! Thank you Greg!

.github/workflows/test.yml
g-k: Yeah, it's just noise for now.

.github/workflows/test.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Install dependencies
+        shell: bash
+        run: |
+          make install install-docs
+
+      - name: Run Tests
+        shell: bash
+        run: |
+          (source venv/bin/activate && make black check_conftest_imports doctest coverage doc-build)
```
hwine, line 45: nit: I don't think the source is needed anymore -- it's addressed in the Makefile as of 2ae329b37f5960b2cb5aa7e677e0e81ef7527bca

hwine (approved):
1 nit

As for testing gh-pages publishing, I did that previously using a fork in my personal account. That wasn't "easy" (lots of configuration points), so I'd suggest that route only if there are problems.

I believe the way to test pypi publishing is via the pypi staging/test server, but have no experience with that.

.github/workflows/test.yml
g-k: dropped!

g-k: hrm failing https://github.com/mozilla/frost/pull/432/checks?check_run_id=1670520524

`make install` makes the venv but the other targets fail in `check_venv`

PR #434, opened 2021-01-08
Fix 429 remove travis

g-k:
fix: #429 

remove travis CI to finish replacing it with GHA from #432 #433 and other commits to the default branch

r? @ajvb


ajvb (approved):
:shipit: 🎉

PR #436, opened 2021-01-21
GCP / GSuite fixes

### LINKED ISSUE #353: Replace deprecated oauth2lib library
The OAuth library in use ([oauth2lib](https://pypi.org/project/oauth2client/4.1.3/)) is depricated, and has not received any updates since [2018-09-20](https://github.com/googleapis/oauth2client)

The suggested replacements are [google-auth](https://google-auth.readthedocs.io/) and [oauthlib](http://oauthlib.readthedocs.io/), but those recommendations are 2 years old.

To do:
- [ ] find a still supported library that is up to snuff
- [ ] integrate it

ajvb:
* Ignore noisy GCP warnings
* Include additional filter for "this service is not enabled" errors within the GCP Client
* Swap out oauth libs for GSuite - Fixes #353
* Remove sum() call from datasets fetching. 
* Add README to gsuite/ explaining that it is experimental.


gcp/client.py
```diff
@@ -1,12 +1,6 @@
-import warnings
-
 from apiclient.discovery import build as build_service
 from apiclient.errors import HttpError
 
-warnings.filterwarnings(
-    "ignore", "Your application has authenticated using end user credentials"
-)
-
```
hwine, line 9: Question: what's the thinking of moving this to `__init__.py`. I.e. I would have thought this code belonged in `client.py` based on our past discussions.

I'm not against it, I just want to understand it so I can use it properly.  Maybe a doc update is needed?

gcp/bigquery/resources.py
```diff
@@ -15,7 +15,7 @@ def datasets():
             get_dataset(d["datasetReference"]["datasetId"], project_id)
             for d in datasets
         ]
-    return sum(results, [])
+    return results
```
hwine, line 18: You're changing the return type here (from int to list) -- but I don't see any changes for where this function is invoked. Are we sure that code will know what to do?

Or maybe add typing to this function and see what mypy has to say?

hwine (approved):
Nice commenting!  Esp where you include the doc URL :grin:

gcp/bigquery/resources.py
ajvb: `sum(var, [])` collapses a list:

```
>>> x = [[1,2,3], ["foo", "bar"]]
>>> x
[[1, 2, 3], ['foo', 'bar']]
>>> sum(x, [])
[1, 2, 3, 'foo', 'bar']
```

This would be needed if we were doing `results.append([get_dataset(d["datasetReference"]["datasetId"], project_id) for d in datasets])`, but we are doing `results += [...]` so it isn't needed.

gcp/client.py
ajvb: I'm open to either. I did worry that moving it to `__init__.py` might be hiding it.

>I.e. I would have thought this code belonged in `client.py` based on our past discussions.

Would you mind reminding me about the past discussion? I don't remember discussing this piece as it relates to which file it should live in.

gcp/bigquery/resources.py
hwine: Ah, so still returns a list. I didn't know about that use of sum, and thinking
```python
>>> sum([1, 2, 3])
6
```

TIL

gcp/client.py
hwine: Roughly (my interpretation):
- `__init__.py` is just to mark directory as a package
- `client.py` has all the code related to communication with the service
- `helpers.py` has all the code related to simplifying tests (e.g. operations on data structures returned from service)

But I see that differs a bit from the [design docs](https://mozilla.github.io/frost/readme-include.html#design).

Again, this isn't a blocker for this commit, but it does look like the design & [file layout](https://mozilla.github.io/frost/readme-include.html#file-layout) needs a refresh.  I'll open a separate issue.

ajvb: >`client.py` has all the code related to communication with the service

Yeah that's a great point. I think it would be best to add these lines back to `client.py`. Will do that now.

